PR #160, opened 2021-08-30
Add dist trainer workflow

aivanou:
Test plan:


aivanou:
this was a test pr to trigger github workflows

PR #176, opened 2021-09-14
specs: allow image to be a Dict[ImageType, str] to allow components to run on multiple schedulers

d4l3k:
This changes image type from `str` to `Union[str, Dict[ImageType, str]]` to allow users to write components that can be seamlessly executed across different environments without having to always override the image.

This is still a WIP and putting it up early to get feedback before fixing all the tests/images.

Here's how the standard library components will look like:
```py
def echo(
    msg: str = "hello world", image: Optional[str] = None, num_replicas: int = 1
) -> specs.AppDef:
    # ...
    return specs.AppDef(
        name="echo",
        roles=[
            specs.Role(
                name="echo",
                image=image or {
                  specs.ImageType.PATH: "/tmp",
                  specs.ImageType.DOCKER: "alpine",
                },
                ...
            )
        ],
    )
```

Image overrides are still specified via strings since that's simplest and during interactive usage users will only be launching on a single scheduler so no reason to specify multiple. If they do want to specify full dict inputs as part of their components that's possible too.

Test plan:

```
$ pytest -n 8
$ pyre
```


d4l3k:
shelving this for now

PR #181, opened 2021-09-16
schedulers/images: add image type detection and warning

d4l3k:
Adds code for detecting image times and warning upfront when trying to schedule a job with an invalid image type.

This changes the local scheduler to detect if the image is a path or a docker image and use the corresponding scheduler. This requires all paths to have a `/` prefix to distinguish them from docker images. This is a bit hacky since it requires the local images to be in a certain format.

Test plan:

```
$ pytest
$ pyre
$ scripts/lint.sh
```

CI


d4l3k:
shelving this for now

PR #219, opened 2021-10-06
Add the skeleton and boilerplate parts of Ray scheduler

cbalioglu:
This is the initial PR for the TorchX Ray scheduler. It is in draft mode since the unit tests are not included yet. Publishing it in advance for communicating the ideas with the rest of the team.

### DIFF
--- a/torchx/schedulers/utils/ray_driver.py
+++ b/torchx/schedulers/utils/ray_driver.py
@@ -20,11 +20,12 @@ class RayActor:
 
 @ray.remote
 class CommandActor:
-    def __init__(self, command):
+    def __init__(self, command, env):
         self.command = command
+        self.env = env
 
     def run_command(self):
-        return subprocess.run(self.command)
+        return subprocess.run(self.command, env = self.env)
 
 class EnhancedJSONEncoder(json.JSONEncoder):
         def default(self, o):
@@ -69,14 +70,9 @@ if __name__ == "__main__":
         logging.debug("Waiting for placement group to start.")
         ready = pg.wait(timeout_seconds=100)
 
+
         if ready:
             logging.debug("Placement group has started.")
-
-            for key, value in actor_dict["env"]:
-                os.environ[key] = value
-
-            logging.debug("Environment variables set")
-
             logging.debug("Starting remote function")
         
         else:
@@ -88,10 +84,10 @@ if __name__ == "__main__":
                 "placement group: {}".format(ray.available_resources(),
                                             pg.bundle_specs))
 
-    actors = [CommandActor.options(placement_group=pgs[i], num_cpus=actor_dict[i]["num_cpus"], num_gpus=actor_dict[i]["num_gpus"]).remote(actors_dict[i]["command"]) for i in range(len(actors_dict) * actor_dict["num_replicas"])]
+    actors = [CommandActor.options(placement_group=pgs[i], num_cpus=actor_dict[i]["num_cpus"], num_gpus=actor_dict[i]["num_gpus"]).remote(actors_dict[i]["command"], actor_dict["env"]) for i in range(len(actors_dict) * actor_dict["num_replicas"])]
     ray.get([a.run_command.remote() for a in actors])
     unfinished = [a.run_command.remote() for a in actors]
-    
+
     while len(unfinished) > 0:
         finished, unfinished = ray.wait(unfinished)
         # If a failure occurs the ObjectRef will be marked as finished.


torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,299 @@
...
+            type_=bool,
+            default=False,
+            help="Copy the script(s) to the cluster. Supported only for PyTorch DDP jobs.",
+        )
+        opts.add(
+            "copy_script_dirs",
+            type_=bool,
+            default=False,
+            help="Copy the directories containing the scripts to the cluster. "
+            "Supported only for PyTorch DDP jobs.",
```
d4l3k, line 128: @aivanou this seems similar to the patch behavior you've been looking to add internally

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,299 @@
...
+        )
+        opts.add(
+            "verbose",
+            type_=bool,
+            default=False,
+            help="Enable verbose output.",
```
d4l3k, line 134: possible to use log levels here instead? is this for the launched job or the scheduler class?

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,299 @@
...
+                    f"The configuration value '{cfg_name}' must be of type {cfg_type.__name__}."
+                )
+
+            setattr(job, cfg_name, cfg_value)
+
+        set_job_attr("cluster_name", str)
+        set_job_attr("start_cluster", bool)
+        set_job_attr("copy_scripts", bool)
+        set_job_attr("copy_script_dirs", bool)
+        set_job_attr("verbose", bool)
```
d4l3k, line 183: generally good to just set the arguments since this is harder to handle with Pyre. Runopts I believe already enforces types (if it doesn't we should probably merge that checkin in there anyways)

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,299 @@
...
+
+            if job.copy_scripts or job.copy_script_dirs:
+                # We support copying scripts to the cluster only for DDP jobs.
+                if (
+                    # fmt: off
+                    actor.command.find(" -m torch.distributed.run ") > 1 or
+                    actor.command.find(" -m torch.distributed.launch ") > 1 or
+                    actor.command.startswith("torchrun")
+                    # fmt: on
+                ):
```
d4l3k, line 214: might be nice to be able to support arbitrary `python foo.py` style commands

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,299 @@
...
+        set_job_attr("copy_script_dirs", bool)
+        set_job_attr("verbose", bool)
+
+        warn_unused_copy_option = True
+
+        for role in app.roles:
+            # Replace the ${img_root}, ${app_id}, and ${replica_id} placeholders
+            # in arguments and environment variables.
+            role = macros.Values(
+                img_root=role.image, app_id=app_id, replica_id="${rank}"
```
d4l3k, line 191: this rank macro is provided by ray?

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,299 @@
...
+        opts.add(
+            "cluster_name",
+            type_=str,
+            help="Override the configured cluster name.",
+        )
+        opts.add(
+            "start_cluster",
+            type_=bool,
+            default=False,
+            help="Start the cluster if needed.",
```
d4l3k, line 115: does ray automatically tear down the cluster after the job finishes or will this scheduler need to do that as well?

msaroufim: Need to call `ray down` https://docs.ray.io/en/latest/cluster/cloud.html

I'll add more detail

torchx/schedulers/ray_scheduler.py
msaroufim: An alternative is setting up a CLI with https://typer.tiangolo.com/ - I'm using it for torchprep and it's an absolute pleasure to use

torchx/schedulers/ray_scheduler.py
msaroufim: Do you do the same in the rest of `torchX` as in make it easy to run arbitrary Python code which may or may not include Pytorch related code

amogkam:
Hey guys, sorry for the delay here- I will take a look at this by EOD tomorrow

PR #262, opened 2021-10-15
Make `output_path` optional

aivanou:
Summary: Since we removed distributed sum, we need to use this example to run fb internal tests. For internal tests, we don't need the `output_path`, which introduces around ~200 mb of data on each run

Reviewed By: kiukchung

Differential Revision: D31661378


d4l3k:
already landed

PR #295, opened 2021-10-21
switch to torch 1.10

d4l3k:
Test plan:

### DIFF
--- a/torchx/runtime/container/Dockerfile
+++ b/torchx/runtime/container/Dockerfile
@@ -2,27 +2,12 @@ FROM pytorch/pytorch:1.9.0-cuda10.2-cudnn7-runtime
 
 RUN pip install fsspec[s3]
 
-# Note: the pytorch image installs version of the torchvision that is not compatible
-# with CUDA. By using it directly there will be a segfault.
-RUN pip uninstall -y torchvision
-RUN pip install torchvision
-
 WORKDIR /app
 
 # copy requirements early so we don't have to redownload dependencies on code
 # changes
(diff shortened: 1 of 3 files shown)


torchx/runtime/container/Dockerfile
```diff
@@ -2,27 +2,12 @@ FROM pytorch/pytorch:1.9.0-cuda10.2-cudnn7-runtime
```
kiukchung, line 2: this would change to `pytorch/pytorch:1.10-cuda10.2-cudnn7-runtime` once that's out on dockerhub.

torchx/runtime/container/Dockerfile
```diff
@@ -2,27 +2,12 @@ FROM pytorch/pytorch:1.9.0-cuda10.2-cudnn7-runtime
...
-# TODO: Remove this and get dependency on torch 1.10 when it is released
-
-RUN pip uninstall -y torch
-
-RUN pip install https://download.pytorch.org/whl/test/cu113/torch-1.10.0%2Bcu113-cp37-cp37m-linux_x86_64.whl
-
-RUN pip uninstall -y torchtext
-
-RUN pip install https://download.pytorch.org/whl/test/torchtext-0.10.0-cp37-cp37m-linux_x86_64.whl
+RUN pip install --upgrade -r dev-requirements.txt
```
kiukchung, line 10: and when we use pytorch/pytorch-1.10 base image we can just do:

```
RUN pip install -r dev-requirements.txt
```
which would skip installing torch-1.10 since its already there in the base image.

PR #304, opened 2021-10-22
local_scheduler: add support for logging stdout via log_iter

d4l3k:
Currently log_iter for local scheduler only logs out stderr. This changes it to log both stderr and stdout as well as adds some better error messages since the current errors are confusing for invalid replica/role name.

Test plan:

```
pytest torchx/schedulers/test/local_scheduler_test.py
pyre
```


d4l3k:
#323 replaces this

PR #315, opened 2021-10-28
Add the skeleton and boilerplate parts of Ray scheduler

cbalioglu:
This is the initial PR of the TorchX Ray scheduler. It mostly contains the boilerplate parts of the scheduler, including the skeleton of the interface methods that should be implemented in the near future.

Note that this PR is a slight revision of #219. I preferred to open a new PR since #219 later grew to contain additional commits that were not meant to be part of the original PR.


setup.py
```diff
@@ -77,6 +77,7 @@ def get_nightly_version():
         extras_require={
             "kfp": ["kfp==1.6.2"],
             "kubernetes": ["kubernetes>=11"],
+            "ray": ["ray>=1.6.0"],
             "dev": dev_reqs,
```
d4l3k, line 81: can you add ray to the dev requirements so these tests run in CI?

torchx/schedulers/__init__.py
```diff
@@ -22,6 +22,17 @@ def __call__(self, session_name: str, **kwargs: object) -> Scheduler:
         ...
 
 
+def add_optional_ray_scheduler_factory(factories: Dict[str, SchedulerFactory]) -> None:
```
d4l3k, line 25: we generally handle missing imports by importing within the relevant functions. Makes it a bit simpler usage wise

i.e.
https://github.com/pytorch/torchx/blob/main/torchx/schedulers/kubernetes_scheduler.py#L361

for top level types you can throw them behind TYPE_CHECKING:
https://github.com/pytorch/torchx/blob/main/torchx/schedulers/kubernetes_scheduler.py#L64

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,237 @@
...
+            The number of GPUs to allocate.
+    """
+
+    name: str
+    command: str
+    env: Dict[str, str] = field(default_factory=dict)
+    num_replicas: int = 1
+    num_cpus: int = 1
+    num_gpus: int = 0
+    # TODO: memory_size, max_retries, retry_policy
```
d4l3k, line 46: Does Ray have native python types for this? If so we should probably use them here. 

Having this class just copy the fields from Role isn't terribly useful. We typically try to match the types that will actually get called via schedule so if the torchx scheduler is missing a field programmatic usage can still add any missing args/fields.

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,237 @@
...
+            type_=bool,
+            default=False,
+            help="Copy the Python script(s) to the cluster.",
+        )
+        opts.add(
+            "copy_script_dirs",
+            type_=bool,
+            default=False,
+            help="Copy the directories containing the Python scripts to the cluster.",
+        )
```
d4l3k, line 123: can we merge this with copy_scripts? copy the dir if it's a dir, copy the file if it's a file?

This also might fit better with a proper canary interface that we've been thinking about adding cc @kiukchung

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,237 @@
...
+        app_id: str,
+        role_name: str,
+        k: int = 0,
+        regex: Optional[str] = None,
+        since: Optional[datetime] = None,
+        until: Optional[datetime] = None,
+        should_tail: bool = False,
+    ) -> Iterable[str]:
+        # TBD
+        return []
```
d4l3k, line 233: would be good to just leave these unimplemented / raise an Unimplemented error for now

torchx/schedulers/test/ray_scheduler_test.py
```diff
@@ -0,0 +1,273 @@
...
+    from torchx.schedulers.ray_scheduler import RayScheduler, logger
+
+    should_skip = False
+except ImportError:
+    should_skip = True
+
+from torchx.specs import AppDef, Resource, Role, RunConfig, runopts
+
+
+@skipIf(should_skip, "Ray is not installed.")
```
d4l3k, line 25: it might be better to wrap these in If statements -- internally it gets annoying since skip tests show up as action items

same way we handle docker https://github.com/pytorch/torchx/blob/main/torchx/schedulers/test/docker_scheduler_test.py#L102

d4l3k (commented):
generally this seems pretty reasonable though doesn't contain the actual meat of the ray integration yet

Are you planning on landing this incrementally? i.e. skeleton in this PR and then the remainder in another one? Either is fine just want to be on the same page since I'm wondering if you're looking for an approval or just incremental feedback

torchx/schedulers/ray_scheduler.py
aivanou: yeah, I think this is just a POC of how to translate torchx.Role to ray.Actor . The actual functionality of scheduling AppDef is still in development. When it will be finalized, I would expect some changes to the class, maybe even removing `RayJob` and `RayActor` in favor of translating them directly into ray API.

torchx/schedulers/__init__.py
cbalioglu: Do you mean within `ray_scheduler.create_scheduler()`?

torchx/schedulers/ray_scheduler.py
cbalioglu: `RayActor` is a "DAO" (a.k.a. Data Access Object) in CRUD terminology. It will be dumped into a JSON file, and read by the `ray_driver.py` on the cluster head node. In a later PR it will be moved to a `ray_common.py` script that will be uploaded to the cluster along with `ray_driver.py`. I avoid using native TorchX types like `Role` or `AppDef` here since that would require us to install TorchX on the cluster and unnecessarily depend on it in the driver script.

torchx/schedulers/ray_scheduler.py
cbalioglu: The user will always specify a file as part of the command line. Unfortunately there is no way for us to figure out what they want in such case. For example:

```
python -m torch.distributed.run ... myscript.py
```

The user has to explicitly specify whether they want us to copy `myscript.py` only or the directory in which it resides.

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,237 @@
...
+
+        for role in app.roles:
+            # Replace the ${img_root}, ${app_id}, and ${replica_id} placeholders
+            # in arguments and environment variables.
+            role = macros.Values(
+                img_root=role.image, app_id=app_id, replica_id="${rank}"
+            ).apply(role)
+
+            actor = RayActor(
+                name=role.name,
```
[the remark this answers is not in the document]
cbalioglu: Replicas of an actor will be created in the driver script on the cluster head node. Here we simply specify the "plain" name (e.g. `worker`, `parameter-server`) and the number of replicas.

torchx/schedulers/__init__.py
cbalioglu: Ok, I think I get what you mean after checking the docker scheduler. I revised the implemented, please check the latest revision.

cbalioglu:
> generally this seems pretty reasonable though doesn't contain the actual meat of the ray integration yet
> 
> Are you planning on landing this incrementally? i.e. skeleton in this PR and then the remainder in another one? Either is fine just want to be on the same page since I'm wondering if you're looking for an approval or just incremental feedback

Yes, this is the skeleton for the Ray scheduler that we want to merge now. The remaining work will be incrementally built and merged on top of this work (mostly by @msaroufim), including the actual Ray integration in `schedule()`, `describe(), `_cancel_existing()`.

torchx/schedulers/ray_scheduler.py
aivanou: The benefit of having a `torchx.Ray*` dataclasses is that we can make all validation during the request creation, and make `ray_driver.py` very dumb(just to get json->create actors), without any validation. This potentially will decrease the time of finding errors

aivanou (approved):

msaroufim (approved):

torchx/schedulers/ray_scheduler.py
```diff
@@ -0,0 +1,236 @@
...
+            The number of GPUs to allocate.
+    """
+
+    name: str
+    command: str
+    env: Dict[str, str] = field(default_factory=dict)
+    num_replicas: int = 1
+    num_cpus: int = 1
+    num_gpus: int = 0
+    # TODO: memory_size, max_retries, retry_policy
```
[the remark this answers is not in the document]
cbalioglu: Let me do it tomorrow as a quick follow-up PR.

PR #329, opened 2021-11-02
Ray scheduler driver and job api

msaroufim:
## To run distributed pytorch test

```
cd torch/torchx/schedulers/test
python -m unittest ray_scheduler_test.py
```

```
2021-11-24 00:53:22,810 INFO worker.py:840 -- Connecting to existing Ray cluster at address: 172.31.2.209:6379
2021-11-24 00:53:23,127 INFO sdk.py:144 -- Uploading package gcs://_ray_pkg_e212ba1ff8e15e24.zip.
2021-11-24 00:53:23,128 INFO packaging.py:352 -- Creating a file package for local directory '/tmp/tmp9copqxd2'.
status: PENDING
status: PENDING
status: RUNNING
status: RUNNING
status: RUNNING
status: RUNNING
status: SUCCEEDED
2021-11-24 00:53:25,521 INFO worker.py:840 -- Connecting to existing Ray cluster at address: 172.31.2.209:6379
(CommandActor pid=3575) initializing `gloo` process group
(CommandActor pid=3574) initializing `gloo` process group
(CommandActor pid=3575) successfully initialized process group
(CommandActor pid=3575) rank: 1, actual world_size: 2, computed world_size: 2
(CommandActor pid=3574) successfully initialized process group
(CommandActor pid=3574) rank: 0, actual world_size: 2, computed world_size: 2
```

## To run distributed pytorch test using torchX CLI

```
# Setup cluster and get a HEAD NODE IP
ray up -y ray_cluster.yaml
pip install torchx[dev]

# Get a job ID from deployed job
torchx run -s ray -cfg dashboard_address=34.209.89.185:20002,working_dir=aivanou_test utils.binary --entrypoint ray_simple.py

# Use job ID to get logs or job status
torchx describe ray://torchx/34.209.89.185:20002-raysubmit_aKvezN3NyA2mqZeW
torchx log ray://torchx/34.209.89.185:20002-raysubmit_aKvezN3NyA2mqZeW
```

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -10,8 +10,8 @@ kfp==1.8.9
 moto==2.2.12
 pyre-extensions==0.0.21
 pytorch-lightning==1.5.6
-ray==1.9.0
 s3fs==2021.10.1
+ray[default]==1.9.2
 torch-model-archiver==0.4.2
 torch==1.10.0
 torchserve==0.4.2
--- a/scripts/component_integration_tests.py
+++ b/scripts/component_integration_tests.py
@@ -37,6 +37,13 @@ def get_k8s_sched_info(image: str) -> SchedulerInfo:
     return SchedulerInfo(name="kubernetes", image=image, cfg=cfg)
 
 
+def get_ray_sched_info(image: str) -> SchedulerInfo:
+    cfg = {
+        "namespace": "torchx-dev",
+    }
+    return SchedulerInfo(name="ray", image=image, cfg=cfg)
+
+
 def get_local_cwd_sched_info(image: str) -> SchedulerInfo:
     return SchedulerInfo(name="local_cwd", image=image)
 
--- a/torchx/components/utils.py
+++ b/torchx/components/utils.py
@@ -147,6 +147,36 @@ def python(
     )
 
 
+def binary(
+    *args: str,
+    entrypoint: str,
+    name: str = "torchx_utils_python",
+    num_replicas: int = 1,
+) -> specs.AppDef:
+    """
+    Test component
+
+    Args:
+        args: arguments passed to the program in sys.argv[1:] (ignored with `--c`)
+        name: name of the job
+        num_replicas: number of copies to run (each on its own container)
+    :return:
+    """
+    return specs.AppDef(
+        name=name,
+        roles=[
+            specs.Role(
+                name="binary",
+                image="<NONE>",
+                entrypoint=entrypoint,
+                num_replicas=num_replicas,
+                resource=specs.Resource(cpu=2, gpu=0, memMB=4096),
+                args=[*args],
+            )
+        ],
+    )
+
+
 def copy(src: str, dst: str, image: str = torchx.IMAGE) -> specs.AppDef:
     """
     copy copies the file from src to dst. src and dst can be any valid fsspec
--- a/torchx/schedulers/__init__.py
+++ b/torchx/schedulers/__init__.py
@@ -5,7 +5,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Dict
+from typing import Dict, Optional
 
 import torchx.schedulers.docker_scheduler as docker_scheduler
 import torchx.schedulers.kubernetes_scheduler as kubernetes_scheduler
@@ -22,6 +22,19 @@ class SchedulerFactory(Protocol):
         ...
 
 
+def try_get_ray_scheduler() -> Optional[SchedulerFactory]:
+    try:
+        from torchx.schedulers.ray_scheduler import _has_ray  # @manual
+
+        if _has_ray:
+            import torchx.schedulers.ray_scheduler as ray_scheduler  # @manual
+
+            return ray_scheduler.create_scheduler
+
+    except ImportError:  # pragma: no cover
+        return None
+
+
 def get_scheduler_factories() -> Dict[str, SchedulerFactory]:
     """
     get_scheduler_factories returns all the available schedulers names and the
@@ -29,12 +42,17 @@ def get_scheduler_factories() -> Dict[str, SchedulerFactory]:
 
     The first scheduler in the dictionary is used as the default scheduler.
     """
+
     default_schedulers: Dict[str, SchedulerFactory] = {
         "local_docker": docker_scheduler.create_scheduler,
         "local_cwd": local_scheduler.create_cwd_scheduler,
         "slurm": slurm_scheduler.create_scheduler,
         "kubernetes": kubernetes_scheduler.create_scheduler,
     }
+
+    ray_scheduler_creator = try_get_ray_scheduler()
+    if ray_scheduler_creator:
+        default_schedulers["ray"] = ray_scheduler_creator
 
     return load_group(
         "torchx.schedulers",
--- a/torchx/schedulers/ray/ray_driver.py
+++ b/torchx/schedulers/ray/ray_driver.py
@@ -1,0 +1,174 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import contextlib
+import importlib
+import json
+import logging
+import os
+import sys
+from typing import Dict, List, Optional, Tuple
+
+import ray
+from ray.train.utils import get_address_and_port
+from ray.util.placement_group import PlacementGroup
+from torchx.schedulers.ray.ray_common import RayActor
+
+_logger: logging.Logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
+
+
+@contextlib.contextmanager
+def redirect_argv(args: List[str]):  # pyre-ignore[3]
+    _argv = sys.argv[:]
+    sys.argv = args
+    yield
+    sys.argv = _argv
+
+
+@ray.remote
+class CommandActor:  # pragma: no cover
+    def __init__(self, command: str, env: Dict[str, str]) -> None:
+        self.args: List[str] = command.split(" ")
+        self.path: str = self.args[0]
+
+        for k, v in env.items():
+            os.environ[k] = v
+
+    def exec_module(self) -> None:
+        spec: Optional[
+            importlib.machinery.ModuleSpec
+        ] = importlib.util.spec_from_file_location("__main__", self.path)
+        if spec:  # pragma: no cover
+            train = importlib.util.module_from_spec(spec)
+            with redirect_argv(self.args):
+                spec.loader.exec_module(train)  # pyre-ignore[16]
+
+    def get_actor_address_and_port(self) -> Tuple[str, int]:
+        return get_address_and_port()
+
+    def set_address_and_port(self, address: str, port: int) -> None:
+        os.environ["MASTER_PORT"] = str(port)
+        os.environ["MASTER_ADDR"] = address
+
+
+def load_actor_json(filename: str) -> List[RayActor]:
+    with open(filename) as f:
+        actors: List[RayActor] = []
+        # Yes this is gross but it works
+        actor_dict = json.load(f)
+        actor_dict = json.loads(actor_dict)
+        for actor in actor_dict:
+            actors.append(RayActor(**actor))
+        return actors
+
+
+def create_placement_groups(actors: List[RayActor]) -> List[PlacementGroup]:
+    pgs: List[PlacementGroup] = []
+    for actor in actors:
+        bundle = {"CPU": actor.num_cpus, "GPU": actor.num_gpus}
+        bundles = [bundle] * actor.num_replicas
+
+        # To change the strategy type
+        # refer to available options here https://docs.ray.io/en/latest/placement-group.html#pgroup-strategy
+        pg = ray.util.placement_group(bundles, strategy="SPREAD")
+        pgs.append(pg)
+
+        _logger.info("Waiting for placement group to start.")
+        ready = pg.wait(timeout_seconds=100)
+
+        if ready:
+            _logger.info("Placement group has started.")
+            _logger.info("Starting remote function")
+        else:  # pragma: no cover
+            raise TimeoutError(
+                "Placement group creation timed out. Make sure "
+                "your cluster either has enough resources or use "
+                "an autoscaling cluster. Current resources "
+                "available: {}, resources requested by the "
+                "placement group: {}".format(ray.available_resources(), pg.bundle_specs)
+            )
+    return pgs
+
+
+def create_command_actors(
+    actors: List[RayActor], pgs: List[PlacementGroup]
+) -> List[CommandActor]:
+
+    # 1. Create actors
+    # 2. For each actor get rank 0 address and port
+    # 3. Set address and port in command actor
+    command_actors: List[CommandActor] = []
+    # address, port = get_address_and_port()
+    for i in range(len(actors)):
+        world_size = actors[i].num_replicas
+        actors_for_this_group = []
+
+        for rank in range(world_size):
+
+            # Environment variables for distributed training
+            rank_env = {
+                "WORLD_SIZE": str(world_size),
+                "RANK": str(rank),
+            }
+
+            actor_and_rank_env = {**actors[i].env, **rank_env}
+
+            actors_for_this_group.append(
+                CommandActor.options(  # pyre-ignore[16]
+                    placement_group=pgs[i],
+                    num_cpus=actors[i].num_cpus,
+                    num_gpus=actors[i].num_gpus,
+                ).remote(actors[i].command, actor_and_rank_env)
+            )
+
+            rank_0_address, rank_0_port = ray.get(
+                actors_for_this_group[0].get_actor_address_and_port.remote()
+            )
+
+            for actor in actors_for_this_group:
+                ray.get(actor.set_address_and_port.remote(rank_0_address, rank_0_port))
+
+            command_actors.extend(actors_for_this_group)
+
+    return command_actors
+
+
+if __name__ == "__main__":  # pragma: no cover
+    _logger.debug("Reading actor.json")
+
+    actors: List[RayActor] = load_actor_json("actors.json")
+    os.remove("actors.json")
+
+    _logger.debug("Creating Ray placement groups")
+    ray.init(address="auto", namespace="torchx-ray")
+    pgs: List[PlacementGroup] = create_placement_groups(actors)
+
+    _logger.debug("Getting command actors")
+    command_actors: List[CommandActor] = create_command_actors(actors, pgs)
+
+    _logger.debug("Running Ray actors")
+    active_workers = [  # pyre-ignore
+        command_actor.exec_module.remote()  # pyre-ignore
+        for command_actor in command_actors
+    ]
+
+    # Await return result of remote ray function
+    while len(active_workers) > 0:
+        completed_workers, active_workers = ray.wait(active_workers)
+        # If a failure occurs the ObjectRef will be marked as completed.
+        # Calling ray.get will expose the failure as a RayActorError.
+        for object_ref in completed_workers:
+            try:
+                ray.get(object_ref)
+                _logger.info("Ray remote function promise succesfully returned")
+
+            # If an error occurs during the actor execution, this error will get propagated as-is to the driver when you call ray.get().
+            # For example, if a ValueError is raised in the actor method call, this will be raised as a ValueError on the driver.
+            # These exceptions will not be caught in this try-except clause
+            except ray.exceptions.RayActorError as exc:
+                _logger.info("Ray Actor Error")
+                _logger.error("Ray Actor error", exc)
--- a/torchx/schedulers/ray_scheduler.py
+++ b/torchx/schedulers/ray_scheduler.py
@@ -4,236 +4,329 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import dataclasses
+import json
 import logging
 import os
+import time
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import Mapping, Any, Dict, Iterable, List, Optional, Set, Type
+from shutil import copy2, rmtree, copytree
+from tempfile import mkdtemp
+from typing import Any, Dict, List, Mapping, Optional, Set, Type
 
-from torchx.schedulers.api import AppDryRunInfo, DescribeAppResponse, Scheduler, Stream
+from torchx.schedulers.api import (
+    AppDryRunInfo,
+    AppState,
+    DescribeAppResponse,
+    Scheduler,
+    Stream,
+)
 from torchx.schedulers.ids import make_unique
-from torchx.specs import AppDef, CfgVal, SchedulerBackend, macros, runopts
-
+from torchx.schedulers.ray.ray_common import RayActor
+from torchx.specs import (
+    AppDef,
+    CfgVal,
+    SchedulerBackend,
+    macros,
+    runopts,
+    Role,
+    RoleStatus,
+    ReplicaStatus,
+)
 
 try:
-    import ray  # @manual # noqa: F401
+    from ray.autoscaler import sdk as ray_autoscaler_sdk
+    from ray.dashboard.modules.job.common import JobStatus
+    from ray.dashboard.modules.job.sdk import JobSubmissionClient
 
     _has_ray = True
+
 except ImportError:
     _has_ray = False
 
+if _has_ray:
 
-_logger: logging.Logger = logging.getLogger(__name__)
+    _logger: logging.Logger = logging.getLogger(__name__)
 
+    _ray_status_to_torchx_appstate: Dict[JobStatus, AppState] = {
+        JobStatus.PENDING: AppState.PENDING,
+        JobStatus.RUNNING: AppState.RUNNING,
+        JobStatus.SUCCEEDED: AppState.SUCCEEDED,
+        JobStatus.FAILED: AppState.FAILED,
+        JobStatus.STOPPED: AppState.CANCELLED,
+    }
 
-def has_ray() -> bool:
-    """Indicates whether Ray is installed in the current Python environment."""
-    return _has_ray
+    class _EnhancedJSONEncoder(json.JSONEncoder):
+        def default(self, o: RayActor):  # pyre-ignore[3]
+            if dataclasses.is_dataclass(o):
+                return dataclasses.asdict(o)
+            return super().default(o)
 
+    def serialize(
+        actors: List[RayActor], dirpath: str, output_filename: str = "actors.json"
+    ) -> None:
+        actors_json = json.dumps(actors, cls=_EnhancedJSONEncoder)
+        with open(os.path.join(dirpath, output_filename), "w") as tmp:
+            json.dump(actors_json, tmp)
 
-@dataclass
-class RayActor:
-    """Describes an actor (a.k.a. role in TorchX terms).
+    def has_ray() -> bool:
+        """Indicates whether Ray is installed in the current Python environment."""
+        return _has_ray
 
-    Attributes:
-        name:
-            The name of the actor.
-        command:
-            The command that the actor should run as a subprocess.
-        env:
-            The environment variables to set before executing the command.
-        num_replicas:
-            The number of replicas (i.e. Ray actors) to run.
-        num_cpus:
-            The number of CPUs to allocate.
-        num_gpus:
-            The number of GPUs to allocate.
-    """
+    @dataclass
+    class RayJob:
+        """Represents a job that should be run on a Ray cluster.
 
-    name: str
-    command: str
-    env: Dict[str, str] = field(default_factory=dict)
-    num_replicas: int = 1
-    num_cpus: int = 1
-    num_gpus: int = 0
-    # TODO: memory_size, max_retries, retry_policy
+        Attributes:
+            app_id:
+                The unique ID of the application (a.k.a. job).
+            cluster_config_file:
+                The Ray cluster configuration file.
+            cluster_name:
+                The cluster name to use.
+            dashboard_address:
+                The existing dashboard IP address to connect to
+            working_dir:
+            The working directory to copy to the cluster
+            requirements:
+                The libraries to install on the cluster per requirements.txt
+            actors:
+                The Ray actors which represent the job to be run. This attribute is
+                dumped to a JSON file and copied to the cluster where `ray_main.py`
+                uses it to initiate the job.
+        """
 
+        app_id: str
+        cluster_config_file: Optional[str] = None
+        cluster_name: Optional[str] = None
+        dashboard_address: Optional[str] = None
+        working_dir: Optional[str] = None
+        requirements: Optional[str] = None
+        scripts: Set[str] = field(default_factory=set)
+        actors: List[RayActor] = field(default_factory=list)
 
-@dataclass
-class RayJob:
-    """Represents a job that should be run on a Ray cluster.
+    class RayScheduler(Scheduler):
+        def __init__(self, session_name: str) -> None:
+            super().__init__("ray", session_name)
 
-    Attributes:
-        app_id:
-            The unique ID of the application (a.k.a. job).
-        cluster_config_file:
-            The Ray cluster configuration file.
-        cluster_name:
-            The cluster name to use.
-        copy_script:
-            A boolean value indicating whether to copy the script files to the
-            cluster.
-        copy_script_dir:
-            A boolean value indicating whether to copy the directories
-            containing the scripts to the cluster.
-        scripts:
-            The set of scripts to copy to the cluster.
-        actors:
-            The Ray actors which represent the job to be run. This attribute is
-            dumped to a JSON file and copied to the cluster where `ray_main.py`
-            uses it to initiate the job.
-        verbose:
-            A boolean value indicating whether to enable verbose output.
-    """
+        # TODO: Add address as a potential CLI argument after writing ray.status() or passing in config file
+        def run_opts(self) -> runopts:
+            opts = runopts()
+            opts.add(
+                "cluster_config_file",
+                type_=str,
+                required=False,
+                help="Use CLUSTER_CONFIG_FILE to access or create the Ray cluster.",
+            )
+            opts.add(
+                "cluster_name",
+                type_=str,
+                help="Override the configured cluster name.",
+            )
+            opts.add(
+                "dashboard_address",
+                type_=str,
+                required=False,
+                default="127.0.0.1:8265",
+                help="Use ray status to get the dashboard address you will submit jobs against",
+            )
+            opts.add(
+                "working_dir",
+                type_=str,
+                help="Copy the the working directory containing the Python scripts to the cluster.",
+            )
+            opts.add("requirements", type_=str, help="Path to requirements.txt")
+            return opts
 
-    app_id: str
-    cluster_config_file: str
-    cluster_name: Optional[str] = None
-    copy_scripts: bool = False
-    copy_script_dirs: bool = False
-    scripts: Set[str] = field(default_factory=set)
-    actors: List[RayActor] = field(default_factory=list)
-    verbose: bool = False
+        def schedule(self, dryrun_info: AppDryRunInfo[RayJob]) -> str:
+            cfg: RayJob = dryrun_info.request
 
+            # Create serialized actors for ray_driver.py
+            actors = cfg.actors
+            dirpath = mkdtemp()
+            serialize(actors, dirpath)
 
-class RayScheduler(Scheduler):
-    def __init__(self, session_name: str) -> None:
-        super().__init__("ray", session_name)
+            ip_address = "127.0.0.1:8265"
 
-    def run_opts(self) -> runopts:
-        opts = runopts()
-        opts.add(
-            "cluster_config_file",
-            type_=str,
-            required=True,
-            help="Use CLUSTER_CONFIG_FILE to access or create the Ray cluster.",
-        )
-        opts.add(
-            "cluster_name",
-            type_=str,
-            help="Override the configured cluster name.",
-        )
-        opts.add(
-            "copy_scripts",
-            type_=bool,
-            default=False,
-            help="Copy the Python script(s) to the cluster.",
-        )
-        opts.add(
-            "copy_script_dirs",
-            type_=bool,
-            default=False,
-            help="Copy the directories containing the Python scripts to the cluster.",
-        )
-        opts.add(
-            "verbose",
-            type_=bool,
-            default=False,
-            help="Enable verbose output.",
-        )
-        return opts
+            if cfg.cluster_config_file:
+                ip_address = ray_autoscaler_sdk.get_head_node_ip(
+                    cfg.cluster_config_file
+                )  # pragma: no cover
+            if cfg.dashboard_address:
+                ip_address = cfg.dashboard_address
 
-    def schedule(self, dryrun_info: AppDryRunInfo[RayJob]) -> str:
-        raise NotImplementedError()
+            # 0. Create Job Client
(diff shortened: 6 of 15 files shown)


torchx/schedulers/utils/ray_driver.py
```diff
@@ -0,0 +1,80 @@
...
+def load_actor_json(filename : str) -> List[Dict]:
+    with open(filename) as f:
+        actor = json.load(f)
+        actor = json.loads(actor)
+        return actor
+
+if __name__ == "__main__":
+
+    # Move to scheduler
+    actor1 = RayActor("resnet", ["echo", "hello resnet"])
```
cbalioglu, line 28: These three lines should be removed from the PR. Any kind of test you want to have, should be placed under `torchx/schedulers/test`.

torchx/schedulers/utils/ray_driver.py
```diff
@@ -0,0 +1,80 @@
...
+        self.command = command
+        self.env = env
+
+    def run_command(self):
+        return subprocess.run(self.command, env = self.env)
+
+def load_actor_json(filename : str) -> List[Dict]:
+    with open(filename) as f:
+        actor = json.load(f)
+        actor = json.loads(actor)
```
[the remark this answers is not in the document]
cbalioglu: If we can use something like `dataclasses-json` as I mentioned earlier, we can return a `RayActor` here, instead of a `dict`. This is also why I suggested moving `RayActor` to `ray_common.py`, so that it can be shared by both `ray_scheduler.py` and `ray_driver.py`.

torchx/schedulers/utils/ray_driver.py
```diff
@@ -0,0 +1,80 @@
...
+
+    for actor_dict in actors_dict:
+
+        bundle = {"CPU": actor_dict["num_cpus"], "GPU": actor_dict["num_gpus"]}
+        bundles = [bundle] * actor_dict["num_replicas"]
+        pg = ray.util.placement_group(bundles, strategy="SPREAD")
+        pgs.append(pg)
+
+        logging.debug("Waiting for placement group to start.")
+        ready = pg.wait(timeout_seconds=100)
```
cbalioglu, line 49: Let's keep the timeout 100 seconds. Once we implement the `timeout` option in `ray_scheduler` we can update this.

torchx/schedulers/utils/ray_driver.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # On driver.py
+    logging.debug("Reading actor.json")
+    actors_dict = load_actor_json('actor.json')
+    pgs = []
+    ray.init(address="auto", namespace="torchx-ray")
+
+    for actor_dict in actors_dict:
+
+        bundle = {"CPU": actor_dict["num_cpus"], "GPU": actor_dict["num_gpus"]}
```
cbalioglu, line 43: I would advise getting @amogkam's feedback on the rest of `ray_driver.py`. We should pay particular attention to the edge cases. How do we handle failures? For instance what happens if an actor or a process group initialization raises an error? How do we communicate the root cause back to the user?

torchx/schedulers/utils/ray_driver.py
```diff
@@ -0,0 +1,80 @@
...
+        # If a failure occurs the ObjectRef will be marked as finished.
+        # Calling ray.get will expose the failure as a RayActorError.
+        for object_ref in finished:
+            try:
+                status_code = ray.get(object_ref)
+            except ray.RayActorError as exc:
+                status_code = 1
+
+            if status_code != 0:
+                raise RuntimeError("Job failed")
```
cbalioglu, line 80: Is this the way we should report a job failure to Ray? Or is there a better approach? Can the user see `stdout` and `stderr` of the subprocesses running inside the remote actors? (Similar to how you see outputs of each process in a Slurm job)

torchx/schedulers/utils/ray_driver.py
```diff
@@ -0,0 +1,80 @@
...
+                "available: {}, resources requested by the "
+                "placement group: {}".format(ray.available_resources(),
+                                            pg.bundle_specs))
+
+    actors = [CommandActor.options(placement_group=pgs[i], num_cpus=actor_dict[i]["num_cpus"], num_gpus=actor_dict[i]["num_gpus"]).remote(actors_dict[i]["command"], actor_dict["env"]) for i in range(len(actors_dict) * actor_dict["num_replicas"])]
+    ray.get([a.run_command.remote() for a in actors])
+    unfinished = [a.run_command.remote() for a in actors]
+
+    while len(unfinished) > 0:
+        finished, unfinished = ray.wait(unfinished)
```
cbalioglu, line 70: Does `ray.wait()` remove finished actors from `unfinished`? How does the condition `len(unfinished) > 0` ever become false?

torchx/schedulers/utils/ray_driver.py
```diff
@@ -0,0 +1,80 @@
...
+    def run_command(self):
+        return subprocess.run(self.command, env = self.env)
+
+def load_actor_json(filename : str) -> List[Dict]:
+    with open(filename) as f:
+        actor = json.load(f)
+        actor = json.loads(actor)
+        return actor
+
+if __name__ == "__main__":
```
cbalioglu, line 25: I suggest moving the rest of the code in this file into a function (e.g. `def _main()`). Otherwise, you won't be able to have proper tests for it.

cbalioglu (changes requested):
I left some comments.

Besides those comments one major thing that is missing right now are the tests. I am not entirely sure to what extent we can test all the logic contained in this PR, but we should give it a try. I suggest syncing up with Aliaksandr and Tristan to learn how they test Slurm, Kubernetes, or Docker schedulers. We should have a similar approach here.

msaroufim:
Chatted briefly with @d4l3k on what an integration test could look like


hat we can do is setup a single node Ray cluster on the machine where the test is running that would instantiate a single Ray node with multiple workers, run the test and tear down the cluster

For multi node we can leverage something like `ray up` and `ray down` to setup an infrastructure but will be a bit more work

cbalioglu:
@amogkam is it possible to simulate a Ray cluster on localhost?

torchx/schedulers/ray_scheduler.py
```diff
@@ -23,6 +24,18 @@
 
 _logger: logging.Logger = logging.getLogger(__name__)
 
+class EnhancedJSONEncoder(json.JSONEncoder):
```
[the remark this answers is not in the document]
msaroufim: I did it has some quirks I haven't worked through yet

At a high level it ends up looking like this 


```
def serialize_lib(actors: List[RayActor], output_filename="actor.json"):
  actors_json = RayActor.schema().dump(actors, many=True)
  with open(f"{output_filename}", 'w', encoding='utf-8') as f:
    json.dump(actor_json, f)
```

d4l3k:
@cbalioglu sorry for that request, accidentally clicked

torchx/schedulers/utils/ray_driver.py
```diff
@@ -0,0 +1,80 @@
...
+    ray.get([a.run_command.remote() for a in actors])
+    unfinished = [a.run_command.remote() for a in actors]
+
+    while len(unfinished) > 0:
+        finished, unfinished = ray.wait(unfinished)
+        # If a failure occurs the ObjectRef will be marked as finished.
+        # Calling ray.get will expose the failure as a RayActorError.
+        for object_ref in finished:
+            try:
+                status_code = ray.get(object_ref)
```
amogkam: If we go with the comment above about not running the command in a subprocess, then the logic for the `status_code` should not be needed. All exceptions raised by the actor will then get automatically propagated to the driver.

torchx/schedulers/utils/ray_driver.py
amogkam: There's a couple points here:

1. For stdout and stderr, if we run the command in the main process instead of the subprocess (see the comment above about using `exec`), then the stdout and stderr from each actor will be shown on the driver side. Then, Ray Jobs will take the stdout and stderr from the driver (which contains the stdout and stderr from each actor) and persists it on the head node. Ray jobs exposes a `/logs` endpoint that torchX (or the user) can use to retrieve these logs. Does that work for you guys?
2. For exceptions, if an actor raises an error during the execution of the command, then this **same exception** will be raised when you do `ray.get(object_ref)`. This will then get propagated up to the driver, and will terminate the job and the exception message will be piped to stderr. Just as above, stderr can be accessed via the `/logs` endpoint that Ray Jobs exposes. 

Using `exec(...)` instead of `subprocess` should make both of the above much simpler.

Also, my fault for not clarifying this before, but a `RayActorError` will only be raised if the the object reference cannot be resolved (for example if the actor is pre-empted or a node goes down). In that case, we should catch the `RayActorError`, and either have some built-in retry logic or raise a more meaningful error

amogkam:
Hey @cbalioglu @msaroufim for testing a Ray cluster we have a fake cluster utility that you can run on a single machine: https://docs.ray.io/en/master/fake-autoscaler.html.

You can either start a fake multi-node cluster programmatically: https://docs.ray.io/en/master/fake-autoscaler.html#using-ray-cluster-utils-autoscalingcluster

@jiaodong is it possible to submit jobs to the fake cluster?

jiaodong:
For testing job submission, as long as we got a running ray cluster with dashboard module running on headnode, we can use it for testing. Currently what we do the most is start a single node ray cluster and submit jobs to the headnode.

There's a WIP PR that adds a bunch of things to job submission with tests passing: https://github.com/ray-project/ray/pull/19860/files you should be able to find some tests examples there for {python api, sdk, http endpoint} by looking at files in `python/ray/tests/test_job_manager.py`and `dashboard/modules/job/tests/test_http_job_server.py`

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,77 @@
...
+from ray_common import RayActor
+
+@ray.remote
+class CommandActor:
+    def __init__(self, command, env):
+        self.command = command
+        self.env = env
+
+    def run_command(self):
+        return exec(object = self.command, globals = self.env)
```
amogkam, line 22: I don't think you can pass in arbitrary commands to `exec`- it can only take in Python statements.

So instead of passing in a `command` to the actor, we can change this to pass in the name of the train script to run. And then you would do `exec(open(self.train_script))`.

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,77 @@
...
+
+def load_actor_json(filename : str) -> List[Dict]:
+    with open(filename) as f:
+        actor = json.load(f)
+        actor = json.loads(actor)
```
amogkam, line 27: Duplicate?

torchx/schedulers/ray_scheduler.py
```diff
@@ -97,6 +100,15 @@ class RayJob:
 class RayScheduler(Scheduler):
     def __init__(self, session_name: str) -> None:
         super().__init__("ray", session_name)
+        # TODO: Users can run ray.status() to figure out the right URL 
+        # But for now Ray team will explore finding a programmatic way of offering this
+        self.client = JobSubmissionClient("http://127.0.0.1:8265")
```
cbalioglu, line 105: Not sure I am following this method body. I assume this will evolve into something that will be part of the `schedule()` method for resolving the URL of the Job REST API, right?

torchx/schedulers/ray_scheduler.py
```diff
@@ -132,7 +144,44 @@ def run_opts(self) -> runopts:
...
-        raise NotImplementedError()
+        # TODO: What to do with cfg
+        cfg: RayJob = dryrun_info.request
+
+        # TODO: figure out how to use
+        # app_id not needed here - ray will return a job_id
+        app_id: str
+
+        # no cluser config is needed since users will call ray up
+        cluster_config_file: str
```
cbalioglu, line 155: @amogkam Do you know how we will be able to retrieve the URL of the Job REST API without using a cluster YAML file? Right now the Ray CLI relies on the provided YAML file to figure out which cluster to use for a command. How is that going to work with the new Job API? My thinking was something along the line:

```
$ ray get-job-server-url <cluster_config_file>
```

following the idiom of the other Ray commands. Ideally with an accompanying API in the Ray Autoscaler SDK (e.g. `get_job_server_url()` that offers the same functionality programmatically.

torchx/schedulers/ray_scheduler.py
```diff
@@ -6,56 +6,59 @@
...
-    name: str
-    command: str
-    env: Dict[str, str] = field(default_factory=dict)
-    num_replicas: int = 1
-    num_cpus: int = 1
-    num_gpus: int = 0
-    # TODO: memory_size, max_retries, retry_policy
+def serialize(actors : List[RayActor], output_filename='actors.json') -> None:
+    actors_json = json.dumps(actors, cls= EnhancedJSONEncoder)
+    with open(os.path.join("./ray", output_filename)) as tmp:
```
cbalioglu, line 56: This should use a temporary directory as I suggested in an earlier revision (i.e. `NamedTemporaryFile`). We should not pollute user's current working directory, in fact we might not even have write access to the directory, so this is prone to fail as well.

torchx/schedulers/ray_scheduler.py
```diff
@@ -132,7 +144,44 @@ def run_opts(self) -> runopts:
...
+
+
+        job_id : str = self.client.submit_job(
+            # we will pack, hash, zip, upload, register working_dir in GCS of ray cluster
+            # and use it to configure your job execution.
+            # TODO: Make this work for windows
+            # TODO: How is ray/train.py called and pointed to 
+            # TODO: How is train.py called in DDP fashion
+            entrypoint="python ray/ray_driver.py",
+            runtime_env={"working_dir" : "./ray"}
```
cbalioglu, line 181: @amogkam my understanding is that there is no way to specify more than one "working_dir" in an environment definition, is that right? The problem I see right now is, since we can technically execute multiple user scripts that can be potentially in different paths on the local file system, ideally we should be able to specify a "working_dir" per actor. Looks like this is not possible right now. Similarly `ray_driver.py` is a separate file that is not in user's working directory. That needs to be copied separately as well. One idea I can think of is to copy the script directories into a temporary directory along with `ray_driver.py` and specify that temporary directory as "working_dir". It would work, but would also require unnecessarily copying things around and complicating the implementation. Any other ideas you can think of?

torchx/schedulers/ray_scheduler.py
```diff
@@ -132,7 +144,44 @@ def run_opts(self) -> runopts:
...
+        cluster_config_file: str
+
+        # can set this in driver script
+        cluster_name: Optional[str] = None
+        
+        # not sure what to do with this
+        verbose: bool = False
+
+        # not needed?
+        cfg.copy_script_dirs = True
```
cbalioglu, line 164: We still need them; otherwise, we can't now which files or directories to specify as part of `runtime_env`.

torchx/schedulers/ray/train.py
```diff
@@ -0,0 +1,54 @@
+# https://pytorch.org/tutorials/beginner/pytorch_with_examples.html
```
cbalioglu, line 1: This file should be under `torchx/schedulers/test/ray`.

torchx/schedulers/ray_scheduler.py
jiaodong: @cbalioglu working_dir can also be configured at task and actor level, some of the examples can be seen in our tests: https://sourcegraph.com/github.com/ray-project/ray/-/blob/python/ray/tests/test_runtime_env_complicated.py?L822-825

For how currently runtime_env works in the path of jobs submission, we support two paths to minimize user copying files around that's already live on latest nightly wheels:
1) Upload user's local working_dir to running ray cluster with unique id, so all subsequent ray.init(), task / actor can use the same files with same working_dir from its runtime_env config.
2) Support a .zip file on s3 as working_dir in case user wants a stable set of files for repeating workload. This can work either as public bucket or private one with credentials set on the host that executes it.

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,76 @@
...
+import subprocess
+from .ray_common import RayActor
+
+@ray.remote
+class CommandActor:
+    def __init__(self, train_script):
+        self.train_script = train_script
+
+    def run_command(self):
+        return exec(open(self.train_script))
```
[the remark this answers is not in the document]
amogkam: @cbalioglu `exec(open(self.train_script))` should work here since will `open` will read the file and convert it to string form that `exec` can now execute.

The benefit of this vs. `os.execve` or `subprocess` is that there's no extra logic needed to handle exceptions or stdout/stderr. Any exception that's raised by the user script will be treated as an exception occurring in the actor itself, since the user script is being run in the same process as the actor. This also means we don't need to worry about exit codes.

torchx/schedulers/ray_scheduler.py
amogkam: Yeah for now I think it might be best to package everything in a single working dir. 

So either 
1. Allow the user to only specify a single directory and expect all files to be in that directory. You would then just need to copy `ray_driver.py` into that directory (probably as a hidden file) and then specify that as the working directory
2. Or go with your suggestion of creating a new tmp directory by copying over all of the user's provided scripts and `ray_driver.py`.

It should be fine to use the same working_dir for every actor since the actual script that needs to be run is specified in the actor config.

torchx/schedulers/ray_scheduler.py
amogkam: The cluster yaml file is what we use for ray to differentiate between different clusters.

We do have something like this already
either `ray get-head-ip <cluster yaml>` or in python `ray.autoscaler.sdk.get_head_ip("my_cluster.yaml")`

torchx/schedulers/ray_scheduler.py
```diff
@@ -187,17 +194,23 @@ def schedule(self, dryrun_info: AppDryRunInfo[RayJob]) -> str:
...
-        copy2("./ray", dirpath)
+        current_directory = os.path.dirname(os.path.abspath(__file__))
+        copy2(os.path.join(current_directory,"ray", "ray_driver.py"), dirpath)
+        copy2(os.path.join(current_directory,"ray", "ray_common.py"), dirpath)
+
 
         # 3. Copy actor.json
-        actor_path = find_file("actor", "/tmp")
-        copy2(actor_path, dirpath)
+        # actor_path = find_file("actor", "/tmp")
```
cbalioglu, line 203: This part is not needed right? You are already copying `actors.json` in `serialize()`.

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,88 @@
...
+    with open(filename) as f:
+        actors = []
+        for actor_dict in json.load(f):
+            actors.append(RayActor(**actor_dict))
+
+        return actors
+
+def _main(job_id = None):
+    print("Reading actor.json")
+    actors_dict = load_actor_json(str.startswith('actor'))
```
cbalioglu, line 38: Not sure what you mean with `str.startswith('actor')` here. This should be `actors.json` right? (Assuming the Ray Job framework sets the current directory to be `working_dir`)

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,88 @@
...
+    pgs = []
+    ray.init(address="auto", namespace="torchx-ray")
+
+    for actor_dict in actors_dict:
+
+        bundle = {"CPU": actor_dict["num_cpus"], "GPU": actor_dict["num_gpus"]}
```
cbalioglu, line 44: Since `load_actor_json` now returns `RayActor` instances, you should change the `actor_dict[]` expressions.

torchx/schedulers/ray_scheduler.py
```diff
@@ -97,20 +112,28 @@ class RayJob:
...
         )
         opts.add(
             "cluster_name",
             type_=str,
             help="Override the configured cluster name.",
         )
+        opts.add(
+            "cluster_address",
+            type_=str,
+            required=False,
```
cbalioglu, line 134: You should have a default value for this option. Ideally `localhost`.

torchx/schedulers/ray_scheduler.py
```diff
@@ -132,7 +155,65 @@ def run_opts(self) -> runopts:
...
+        serialize(actors, dirpath)
+        
+        if cfg.cluster_config_file:
+            ip_address = ray_autoscaler_sdk.get_head_node_ip(cfg.cluster_config_file)
+
+        elif cfg.cluster_address:
+            ip_address = cfg.cluster_address
+
+        else:
+            ip_address = "127.0.0.1"
```
cbalioglu, line 179: If you define a default value as part of `runopts`, you won't need this last statement.

msaroufim:
Adding final todos here
* Add type annotations to make pyre happy
* Figure out why ray driver not found by ray job scheduler
* See a log message message with pytorch logs in `python -m unittest ray_scheduler_test.py`
* Cleanup and error messages

msaroufim:
Feedback from Alex 11/23
1. Rename cluster address to Ray Dashboard address
2. Remove ray start and ray stop commands replace with ETCD test fixtures. Setup a RayHeadServer class in test class to make things more reusable
3. Clarify which ports do users need to know about - do they only need to know the dashboard port? Make dashboard port configurable, do not hardcode
4. Copy script and script feel redundant - use `fsspec`  instead to work with arbitrary file systems. Copy2 fails with symlinks and same file errors
5. Combine Ray cluster address and Ray yaml config - either http file or ray yaml. Parse as pair of port and host if fail treat as file otherwise port and host
6. Test for existence of ray driver and ray common
7. Take runtime_env with pip and create a class for Image in torchX https://docs.ray.io/en/master/ray-job-submission/overview.html#example-setup and expose to user via YAML

msaroufim:
Feedback from Alex 11/23

PO
1. Create a Ray YAML and showcase how users will use it with enable autoscaling 
2. Run a DDP job https://www.internalfb.com/code/fbsource/[de9078c0d0d58430028634091a78a06836d2699f]/fbcode/torchx/fb/example/component.py - convert ray pid to rank and then set environment variables including rank, world size, master address, master port

P1
1. Rename cluster address to Ray Dashboard address
2. Remove ray start and ray stop commands replace with ETCD test fixtures. Setup a RayHeadServer class in test class to make things more reusable
3. Clarify which ports do users need to know about - do they only need to know the dashboard port? Make dashboard port configurable, do not hardcode
4. Copy script and script feel redundant - use `fsspec`  instead to work with arbitrary file systems. Copy2 fails with symlinks and same file errors
5. Combine Ray cluster address and Ray yaml config - either http file or ray yaml. Parse as pair of port and host if fail treat as file otherwise port and host
6. Test for existence of ray driver and ray common
7. Take runtime_env with pip and create a class for Image in torchX https://docs.ray.io/en/master/ray-job-submission/overview.html#example-setup and expose to user via YAML or ask when requirements.txt will be supported in Ray

d4l3k:
@msaroufim can you rebase this diff so the integ tests run?

msaroufim:
Component test

`ray up cluster.yaml`

`cat cluster.yaml

# An unique identifier for the head node and workers of this cluster.
cluster_name: minimal

# The maximum number of workers nodes to launch in addition to the head
# node. min_workers default to 0.
max_workers: 1

# Cloud-provider specific configuration.
provider:
    type: aws
    region: us-west-2
    availability_zone: us-west-2a

# How Ray will authenticate with newly launched nodes.
auth:
    ssh_user: ubuntu
```

File: aivanou_test/ray_simple.py [hunk 24]
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
aivanou, line 1: lets remove this file

dev-requirements.txt
```diff
@@ -10,8 +10,8 @@ kfp==1.8.9
 moto==2.2.12
 pyre-extensions==0.0.21
 pytorch-lightning==1.5.6
-ray==1.9.0
 s3fs==2021.10.1
+ray[default]
```
aivanou, line 14: can we pin the version here? we had a problem that it takes really huge time(40 minutes) to resolve dependencies for pip if versions are not pinned.

ray_cluster.yaml
```diff
@@ -0,0 +1,142 @@
+# An unique identifier for the head node and workers of this cluster.
```
aivanou, line 1: should we move this file to examples? e.g. torchx/examples/apps/aws/ray ?

scripts/component_integration_tests.py
```diff
@@ -37,6 +37,14 @@ def get_k8s_sched_info(image: str) -> SchedulerInfo:
     return SchedulerInfo(name="kubernetes", image=image, cfg=cfg)
 
 
+def get_ray_sched_info(image: str) -> SchedulerInfo:
+    cfg = {
+        "namespace": "torchx-dev",
+        "queue": "default",
```
aivanou, line 43: what is queue for ray?

torchx/components/utils.py
```diff
@@ -147,6 +147,36 @@ def python(
     )
 
 
+def binary(
```
aivanou, line 150: if this is a test component, lets move it to the tests dir?

torchx/schedulers/__init__.py
```diff
@@ -22,20 +22,38 @@ def __call__(self, session_name: str, **kwargs: object) -> Scheduler:
         ...
 
 
+def try_get_ray_scheduler():  # pyre-ignore[3]
```
aivanou, line 25: do you need here both import try-catch and has-ray function? Can we remove has_ray?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+        self.path: str = self.args[0]
+
+        for k, v in env.items():
+            os.environ[k] = v
+
+    def run_command(self) -> None:
+        spec: Optional[
+            importlib.machinery.ModuleSpec
+        ] = importlib.util.spec_from_file_location("__main__", self.path)
+        if spec:  # pragma: no cover
```
aivanou, line 45: should we raise exception if spec is none?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+        self.args: List[str] = command.split(" ")
+        self.path: str = self.args[0]
+
+        for k, v in env.items():
+            os.environ[k] = v
+
+    def run_command(self) -> None:
+        spec: Optional[
+            importlib.machinery.ModuleSpec
+        ] = importlib.util.spec_from_file_location("__main__", self.path)
```
aivanou, line 44: curious, why are we using `_main_` here, what if we are using something else, will it not work?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+        if spec:  # pragma: no cover
+            train = importlib.util.module_from_spec(spec)
```
aivanou, line 46: lets rename this to something else, e.g. `exec_module`

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+            with redirect_argv(self.args):
+                spec.loader.exec_module(train)  # pyre-ignore[16]
+
+
+def load_actor_json(filename: str) -> List[RayActor]:
+    with open(filename) as f:
+        actors: List[RayActor] = []
+        # Yes this is gross but it works
+        actor_dict = json.load(f)
+        actor_dict = json.loads(actor_dict)
```
aivanou, line 56: this works because your file is double-json encoded, e.g.:
```
# not real code
with open(filename) as f:
  f.write(json.write(json.write(my_obj)))
```

Why is it the case?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+            actors.append(RayActor(**actor))
+        return actors
+
+
+def create_placement_groups(actors: List[RayActor]) -> List[PlacementGroup]:
+    pgs: List[PlacementGroup] = []
+    for actor in actors:
+        bundle = {"CPU": actor.num_cpus, "GPU": actor.num_gpus}
+        bundles = [bundle] * actor.num_replicas
+        pg = ray.util.placement_group(bundles, strategy="SPREAD")
```
aivanou, line 67: curious, why strategy is hardcoded to spread? Should we leave a comment about that?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+                "placement group: {}".format(ray.available_resources(), pg.bundle_specs)
+            )
+    return pgs
+
+
+def create_command_actors(
+    actors: List[RayActor], pgs: List[PlacementGroup]
+) -> List[CommandActor]:
+    command_actors: List[CommandActor] = []
+    address, port = get_address_and_port()
```
aivanou, line 91: what this will return for distributed job, a master node address?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+                    placement_group=pgs[i],
+                    num_cpus=actors[i].num_cpus,
+                    num_gpus=actors[i].num_gpus,
+                ).remote(actors[i].command, actor_and_rank_env)
+            )
+    return command_actors
+
+
+if __name__ == "__main__":  # pragma: no cover
+    _logger.info("Reading actor.json")
```
aivanou, line 118: logger.debug

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+    actors: List[RayActor] = load_actor_json("actors.json")
+
+    _logger.info("Creating Ray placement groups")
```
aivanou, line 121: logger.debug

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+    ray.init(address="auto", namespace="torchx-ray")
+    pgs: List[PlacementGroup] = create_placement_groups(actors)
+
+    _logger.info("Getting command actors")
+    command_actors: List[CommandActor] = create_command_actors(actors, pgs)
+
+    _logger.info("Running Ray actors")
```
aivanou, line 128: logger.debug

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+
+if __name__ == "__main__":  # pragma: no cover
+    _logger.info("Reading actor.json")
+    actors: List[RayActor] = load_actor_json("actors.json")
+
+    _logger.info("Creating Ray placement groups")
+    ray.init(address="auto", namespace="torchx-ray")
+    pgs: List[PlacementGroup] = create_placement_groups(actors)
+
+    _logger.info("Getting command actors")
```
aivanou, line 125: logger.debug

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+    command_actors: List[CommandActor] = create_command_actors(actors, pgs)
+
+    _logger.info("Running Ray actors")
+    unfinished = [  # pyre-ignore
```
aivanou, line 129: lets rename this to `active_workers`

torchx/schedulers/ray/ray_driver.py
```diff
@@ -0,0 +1,144 @@
...
+    while len(unfinished) > 0:
+        finished, unfinished = ray.wait(unfinished)
+        # If a failure occurs the ObjectRef will be marked as finished.
+        # Calling ray.get will expose the failure as a RayActorError.
+        for object_ref in finished:
+            try:
+                ray.get(object_ref)
+                _logger.info("Ray remote function promise succesfully returned")
+            except ray.exceptions.RayActorError as exc:
+                _logger.info("Ray Actor Error")
```
aivanou, line 144: can we log exception as well? logger.error(..., exc)

aivanou: also we need to aggregate this and print exceptions for all failed actors

torchx/schedulers/test/ray_scheduler_test.py
```diff
@@ -266,3 +253,106 @@ def test_submit_dryrun_constructs_actor_command(self) -> None:
...
                 job.actors[0].command, "dummy_entrypoint1 arg1 dummy1.py arg2"
             )
+
+    class RayClusterSetup:
+        _instance = None  # pyre-ignore[4]
+
+        def __new__(cls):  # pyre-ignore[3]
+            if cls._instance is None:
+                cls._instance = super(RayClusterSetup, cls).__new__(cls)
+                ray.shutdown()
```
aivanou, line 263: if this is only to make sure that there was no active ray execution, lets just remove this. 
Is it possible to start ray on custom port? - if yes, lets just start it on custom port

torchx/schedulers/test/ray_scheduler_test.py
```diff
@@ -266,3 +253,106 @@ def test_submit_dryrun_constructs_actor_command(self) -> None:
...
+                os.system("ray start --head")
```
aivanou, line 264: lets read exit code and fail on error?

torchx/schedulers/test/ray_scheduler_test.py
```diff
@@ -266,3 +253,106 @@ def test_submit_dryrun_constructs_actor_command(self) -> None:
...
+                name="ddp",
+                num_cpus=2,
+                num_replicas=2,
+                command=os.path.join(current_dir, "train.py"),
+            )
+
+            ray_job = RayJob(
+                app_id=app_id,
+                dashboard_address="127.0.0.1:8265",
+                # requirements=os.path.join(current_dir, "..", "..", "..", "requirements.txt"),
```
aivanou, line 341: nit: lets remove this?

torchx/schedulers/ray_scheduler.py
```diff
@@ -4,236 +4,323 @@
...
+
+        def schedule(self, dryrun_info: AppDryRunInfo[RayJob]) -> str:
+            cfg: RayJob = dryrun_info.request
+
+            # Create serialized actors for ray_driver.py
+            actors = cfg.actors
+            dirpath = mkdtemp()
+            serialize(actors, dirpath)
+
+            ip_address = "127.0.0.1:8265"
```
aivanou, line 150: remove this?

torchx/schedulers/ray_scheduler.py
```diff
@@ -4,236 +4,323 @@
...
+                        reqs.append(line.strip())
+
+            job_id: str = client.submit_job(
+                # we will pack, hash, zip, upload, register working_dir in GCS of ray cluster
+                # and use it to configure your job execution.
+                entrypoint="python ray_driver.py",
+                runtime_env={"working_dir": dirpath, "pip": reqs},
+                # job_id = cfg.app_id
+            )
+            rmtree(dirpath)
```
aivanou, line 185: lets add try/catch/finally in this method, and do `rmtree` in a finally block

torchx/schedulers/ray_scheduler.py
```diff
@@ -4,236 +4,323 @@
...
+            addr, app_id = app_id.split("-")
+            client = JobSubmissionClient(f"http://{addr}")
+            logs = client.get_job_logs(app_id)
+            client.stop_job(app_id)
+
+        def describe(self, app_id: str) -> Optional[DescribeAppResponse]:
+            addr, app_id = app_id.split("-")
+            client = JobSubmissionClient(f"http://{addr}")
+            status = client.get_job_status(app_id).status
+            print(f"Status is {status}")
```
aivanou, line 288: lets remove this or move to logger.debug?

torchx/schedulers/ray_scheduler.py
msaroufim: this is primarily to make pyre happy since only one of the two args are necessary: dashboard_url or cluster_yaml but pyre doesn't know which

torchx/schedulers/ray/ray_driver.py
msaroufim:
```
    def serialize(
        actors: List[RayActor], dirpath: str, output_filename: str = "actors.json"
    ) -> None:
        # Takes List of Ray Actors dataclasses and turn them into JSON
        actors_json = json.dumps(actors, cls=_EnhancedJSONEncoder)


        with open(os.path.join(dirpath, output_filename), "w") as tmp:
                # Turn the list into a JSON
                json.dump(actors_json, tmp)

```

torchx/schedulers/ray/ray_driver.py
msaroufim: From the Ray source code

```python
def get_address_and_port() -> Tuple[str, int]:
    """Returns the IP address and a free port on this node."""

```



```python
master_addr, master_port = worker_group.execute_single(0, get_address_and_port)
```

Question to @amogkam, would it be preferable here to use this like here https://docs.ray.io/en/master/_modules/ray/train/torch.html

torchx/schedulers/test/ray_scheduler_test.py
msaroufim: @amogkam curious to hear your take on this. Right now I'm using a singleton class to make sure that a Ray cluster gets correctly started and torn down. Then each test which needs to use this cluster would issue a decrement counter and when it hits 0 then cluster starts getting torn down. Is this how you do things in your tests?

torchx/schedulers/ray/ray_driver.py
msaroufim: > also we need to aggregate this and print exceptions for all failed actors

Do you have a favorite way of doing this? I'm thinking of just adding all the exceptions to a list

dev-requirements.txt
jiaodong: Latest ray release is 1.9.2 https://pypi.org/project/ray/ 

Given ray 1.10 is coming out soon i would suggest updating this as soon as it come out, since it has log streaming and request header / cookie support for auth.

torchx/schedulers/test/ray_scheduler_test.py
amogkam: @msaroufim just to confirm- if you start and shutdown a ray cluster for each test individually instead of this singleton pattern, you run into this issue, is that correct: https://github.com/ray-project/ray/issues/9546?

For all of our tests we use `pytest` and not `unittest` so we have not run into this problem.

amogkam: Here's the pytest fixture that we use for testing Ray Jobs.

```python
@pytest.fixture
def call_ray_start(request):
    parameter = getattr(
        request, "param", "ray start --head --num-cpus=1 --min-worker-port=0 "
        "--max-worker-port=0 --port 0")
    command_args = parameter.split(" ")
    out = ray._private.utils.decode(
        subprocess.check_output(command_args, stderr=subprocess.STDOUT))
    # Get the redis address from the output.
    redis_substring_prefix = "--address='"
    address_location = (
        out.find(redis_substring_prefix) + len(redis_substring_prefix))
    address = out[address_location:]
    address = address.split("'")[0]

    yield address

    # Disconnect from the Ray cluster.
    ray.shutdown()
    # Kill the Ray cluster.
    subprocess.check_call(["ray", "stop"])
```

Might be switching back to what it was before the singleton pattern, but using `subprocess` to call ray start and ray stop instead of `os.system`.

dev-requirements.txt
msaroufim: Soundds good

torchx/schedulers/test/ray_scheduler_test.py
msaroufim: Ok this is good to know, I think I'll keep the test as is for now since it works just fine and will revisit this

amogkam: Synced with @msaroufim about this offline- the problem is that the tests can run in parallel, so there could be race conditions if you start and stop the ray cluster in each test individually. Singleton pattern was used to have a global setup and teardown to avoid this issue.

cbalioglu (approved):
LGTM

PR #379, opened 2022-02-01
torchx/runner: log events to torch.monitor

d4l3k:
Summary:
This logs the `torchx.runner.events.Events` to `torch.monitor` as well as the existing event handlers. Once monitor is stable the existing ones will be removed entirely in favor of the new interface.

`torch.monitor` is only available with pytorch 1.11 (or main) so it's a no-op if it's not available.

Differential Revision: D33928333

### DIFF
--- a/torchx/runner/events/__init__.py
+++ b/torchx/runner/events/__init__.py
@@ -59,6 +59,15 @@ def _get_or_create_logger(destination: str = "null") -> logging.Logger:
 def record(event: TorchxEvent, destination: str = "null") -> None:
     _get_or_create_logger(destination).info(event.serialize())
 
+    if destination != "console":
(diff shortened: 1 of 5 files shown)


torchx/runner/events/__init__.py
```diff
@@ -59,6 +59,15 @@ def _get_or_create_logger(destination: str = "null") -> logging.Logger:
 def record(event: TorchxEvent, destination: str = "null") -> None:
     _get_or_create_logger(destination).info(event.serialize())
 
+    if destination != "console":
```
edward-io, line 62: would it make sense to have a console event handler, so we don't need this check?

edward-io (commented):
LGTM, will leave it to torchx people to do the final review.

PR #382, opened 2022-02-07
Add log_path to TorchXRunner

ldworkin:
Summary: Currently we have to use a fixed parameter to specify "log_path" (see https://colab.research.google.com/drive/1kOKw5ddweQGmtyADoXfp-xoVdqvRBWV_?usp=sharing#scrollTo=kH0XT8nBJwkC). This way we'll be able to pass it to the TorchXRunner directly, which is more natural. See test plan in notebook.

Differential Revision: D33979622


kiukchung:
The corresponding internal change (D33979622) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

Issue #443, opened 2022-03-31
[torchx/k8s] support device mounts as device plugin + resource limits


kiukchung:
## Description
Instead of supporting device mounts as volume mounts under privileged mode, support it as a device plugin that needs to be installed on the k8s cluster and add a way to specify resource/limits.

## Motivation/Background
On k8s the docker equivalent of `--device` is supported as [k8s device plugin](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/). This is to support mounting the host's `/dev/**` on the container (just regular bind/volume mount won't work unless under privileged mode which not so great from a security standpoint).

TorchX exposes these device mounts as `Role.mounts` of type `DeviceMount` (see `torchx/specs/api.py`) and for consistency sake supports device mounts on k8s via privileged volume mounts (but as mentioned before has security implications).

## Detailed Proposal

For kubernetes scheduler change the current device mount logic from privileged volume mount to:

1. (prerequisite) have the user install the k8s device plugin for the accelerator they want to device mount
2.  expose `containers/resource/limit` as the [Role.metadata](https://github.com/pytorch/torchx/blob/main/torchx/specs/api.py#L294) under some key (e.g. "k8s.resource.limits")

Basically the resource limit that would be specified in the specs.yaml:

```
---
apiVersion: v1
kind: Pod
metadata:
  name: demo-pod
spec:
  containers:
    - name: demo-container-1
      image: k8s.gcr.io/pause:2.0
      resources:
        limits:
          hardware-vendor.example/foo: 2
```

For the device plugin for the device `example/foo` from `hardware-vendor` would be specified in the `Role` as

```
Role.metadata = {
    "k8s.container.resource.limits:hardware-vendor.example/foo" : 2,
}
```

NOTE: the "key" does not necessarily have to be `k8s.container.resource.limits:hardware-vendor.example/foo`, the implementor can figure out a better scheme.

## Alternatives

The resource/limit can be read off of any of:

1. `Role.metadata` (shown above)
2. `Role.resource.capabilities`
3. `DeviceMount.kubernetes_resource_limits` : this would be an added parameter to DeviceMount which can be configured optionally to support the same device mount on multiple schedulers (and keeps everything tidy in one place).  

Example usage of option 3 would be:

```
Role.mounts = [
  DeviceMount(
    # src and dst paths as usual
    src_path="/dev/infiniband/uverbs0", 
    dst_path="/dev/infiniband/uverbs0", 
[...]

PR #447, opened 2022-04-05
Support k8s device plugins/resource limits (#443)

travistagh:
Added "devices" list of (str, int) tuples to role/resource

Added devices.py to map from named devices to DeviceMounts

Added logic in kubernetes_scheduler to add devices from resource to resource limits

Added logic in aws_batch_scheduler and docker_scheduler to add DeviceMounts for any devices from resource

Test plan:
Unit tests in kubernetes_scheduler, aws_batch_scheduler, and docker_scheduler

### DIFF
--- a/torchx/schedulers/devices.py
+++ b/torchx/schedulers/devices.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+import warnings
+from typing import Mapping, Callable, List, Dict
+
+from torchx.specs.api import DeviceMount
+
+
+def efa_to_devicemounts(num_devices: int) -> List[DeviceMount]:
+    device_mounts = []
+    for device_index in range(0, num_devices):
+        device_mounts.append(
+            DeviceMount(
+                src_path="/dev/infiniband/uverbs" + str(device_index),
+                dst_path="/dev/infiniband/uverbs" + str(device_index),
+            )
+        )
+    return device_mounts
+
+
+DEVICES: Mapping[str, Callable[[int], List[DeviceMount]]] = {
+    "vpc.amazonaws.com/efa": efa_to_devicemounts,
+}
+
+
+def get_device_mounts(devices: Dict[str, int]) -> List[DeviceMount]:
+    """
+    Takes in a list of named devices/quantities, and returns a list of DeviceMount objects
+    based on the mappings defined in DEVICES
+    """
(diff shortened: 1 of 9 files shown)


torchx/schedulers/devices.py
```diff
@@ -0,0 +1,30 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from typing import Mapping, Tuple, Callable
+
+from torchx.specs.api import DeviceMount
+
+DEVICES: Mapping[str, Callable] = {
+    "vpc.amazonaws.com/efa": lambda device_index: DeviceMount(
```
d4l3k, line 12: from our earlier chat we should make this take in # devices

torchx/schedulers/devices.py
```diff
@@ -0,0 +1,30 @@
...
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from typing import Mapping, Tuple, Callable
+
+from torchx.specs.api import DeviceMount
+
+DEVICES: Mapping[str, Callable] = {
```
d4l3k, line 11: would be good to type the Callable, i.e. `Callable[[int], List[DeviceMount]]`

torchx/schedulers/devices.py
```diff
@@ -0,0 +1,30 @@
...
+}
+
+
+def get_device_mounts(devices: Tuple[str, int]) -> list[DeviceMount]:
+    """
+    Takes in a list of named devices/quantities, and returns a list of DeviceMount objects
+    based on the mappings defined in DEVICES
+    """
+    device_mounts = []
+    for device in devices:
```
d4l3k, line 25: can just do: `for device_name, num_devices in devices:`

torchx/schedulers/devices.py
```diff
@@ -0,0 +1,30 @@
...
+        [device_name, num_devices] = device
+        if device_name not in DEVICES:
+            continue
```
d4l3k, line 28: can just throw an error or at least a warning to make sure users don't accidentally fatfinger something

torchx/specs/api.py
```diff
@@ -61,6 +62,7 @@ class Resource:
     cpu: int
     gpu: int
     memMB: int
+    devices: list[Tuple[str, int]] = field(default_factory=list)
```
d4l3k, line 65: Any reason for the list? Using a map makes it so keys are forced to be unique -- we probably don't ever want someone to specify the same resource key twice

d4l3k (commented):
Looking good, thanks for the fast work!

PR #465, opened 2022-04-21
Try changing search strategy

stroxler:
This should only be merged if CI comes back clean.

I'm seeing what happens if we change the search strategy for pyre, which will narrow my search for why pyre is failing to open a file on trunk.


stroxler:
For the moment, using `"none"` is a better strategy

PR #466, opened 2022-04-21
Try setting pyre's site-packages search strategy to "none"

stroxler:
Again, making a PR to help me understand why pyre is failing and narrow my search for where we fail to read a file when trying to type check trunk.

I think that no site package search strategy should actually work okay here - we're explicitly putting site-packages on the search paths list, which makes me think we don't need site-packages handling.

### DIFF
--- a/.pyre_configuration
+++ b/.pyre_configuration
@@ -6,7 +6,7 @@
     ".*/IPython/core/tests/nonascii.*",
     ".*/torchx/examples/apps/compute_world_size/.*"
   ],
-  "site_package_search_strategy": "all",
+  "site_package_search_strategy": "none",
   "source_directories": [
     "typestubs",
     "."


stroxler:
This confirms the bug - the user site packages directory does not exist.

There's a bug in the new pyre feature where we fail to check for existence, assuming that python's `sys` module isn't going to hand us a nonexistant path.

.pyre_configuration
```diff
@@ -6,7 +6,7 @@
     ".*/IPython/core/tests/nonascii.*",
     ".*/torchx/examples/apps/compute_world_size/.*"
   ],
-  "site_package_search_strategy": "all",
+  "site_package_search_strategy": "none",
```
stroxler, line 9: Because we are passing the site packages directory via `--source-paths` on the commandline, we don't want to use any pyre site package search strategy at all at this time.

Once I fix the bug, we can flip this back and get rid of that commandline flag

PR #468, opened 2022-04-21
No search path

stroxler:
With --noninteractive flag set, let's see what happens if we keep using the site packages search strategy of "all" but we don't set the search path.

This is the recommended way to run pyre, assuming it works correctly (but it is also new so we may discover a bug)

### DIFF
--- a/scripts/pyre.sh
+++ b/scripts/pyre.sh
@@ -7,6 +7,5 @@
 
 set -eux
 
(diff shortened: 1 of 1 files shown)


stroxler:
This confirms that the problem is almost certainly a nonexistent user site package.

We're assuming that directory exists in the python side, which is leading the backend to try to list a directory that does not exist.

PR #469, opened 2022-04-22
torch.x -- document dict argument parsing failure case when value contains "="

dracifer:
Summary:
in some cases, the value can contain "=". For example, some base64 encoded values as environment variable. torch.x parser currently does not support this case.

adding a test to capture this failure. wonder if we should consider support this. An example use case can be parsing temp AWS token credential to ddp

also correct a documentation copy-pasta

Reviewed By: kiukchung

Differential Revision: D35560931


kiukchung:
This landed via D35560931 (Apr 2022); closing this stale sync mirror.

PR #472, opened 2022-04-25
Add env, max_retries and mounts arguments to utils.sh

hakanardo:
Summary:
The added args are copied from dist.ddp and works in the
exact same way.


d4l3k:
Thanks!

PR #476, opened 2022-04-28
Support for .dockerignore files used as include lists

hakanardo:
A common pattern is to use .dockerignore files as include-lists by
first ignoring * and then listing whats to be included. This patch
fixes some behavioural differences between how .dockerignore files
are interpreted by torchx and docker, that can appear in such cases.

Test plan:
Added unittests.

### DIFF
--- a/torchx/runner/api.py
+++ b/torchx/runner/api.py
@@ -22,11 +22,11 @@ from torchx.specs import (
     AppHandle,
     AppStatus,
     CfgVal,
-    UnknownAppException,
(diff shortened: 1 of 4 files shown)


d4l3k:
Looks good, thanks! I fixed the pyre error

d4l3k:
@hakanardo I added you to the #torchx PyTorch slack channel if you have any other questions etc. I'd be interested in hearing more about how you're using it and any feature requests etc

PR #488, opened 2022-05-17
ray: fixes for distributed training + cancel command

d4l3k:
This fixes the RayScheduler so distributed jobs using the normal `dist.ddp` work correctly. This redid how the actors where scheduled and how arguments are handled for correctness.

This also adds a `torchx cancel` command since it's handy for killing stuck jobs

* Adds in app name to job ID
* Fixes multiple workers being launched
* Fixes role handling to correctly have replica ID and set rank0_env correctly
* Explicitly uses python3 instead of python to avoid system python2.7

Test plan:

```
torchx run -s ray -c working_dir=.,requirements=./requirements.txt  --wait --log dist.ddp --env LOGLEVEL=INFO -j 2x1 -m scripts.compute_world_size
```

### DIFF
--- a/torchx/cli/cmd_cancel.py
+++ b/torchx/cli/cmd_cancel.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import argparse
+import logging
+
+from torchx.cli.cmd_base import SubCommand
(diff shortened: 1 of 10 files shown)


torchx/cli/cmd_cancel.py
```diff
@@ -0,0 +1,33 @@
...
+import sys
+
+from torchx.cli.cmd_base import SubCommand
+from torchx.runner import get_runner
+from torchx.specs import api
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class CmdCancel(SubCommand):
```
colin2328, line 21: nice, thanks

Issue #370, opened 2022-01-21
[torchx/unittest] Make it easy to run components from a unittest


kiukchung:
## Description
Offer a test fixture to make it easy to write unittests that run existing components.

## Motivation/Background
Currently, it is easy to run a component locally via the CLI:
```
$ torchx run $COMPONENT_NAME $COMPONENT_ARGS
```

But not so much programmatically (`foo()` is a component):
```
import os
import unittest
import torchx
from my.components import foo

class FooTest(unittest.TestCase):

    def setUp(self) -> None:
        self.test_dir = tempfile.mkdtemp()
        # this isn't strictly required for OSS but it is for Meta and all of our own tests 
        # (b/c they run in sandcastle)
        self.old_cwd = os.getcwd()
        os.chdir(os.path.dirname(__file__))
        
        self.runner = torchx.get_runner()

    def tearDown(self) -> None:

    def foo_test(self) -> None:
         app_handle = self.runner.run(foo(args...), scheduler="local_cwd", cfg={"prepend_cwd": True, "log_dir": self.test_dir})
         elapsed = 0
         status = self.runner.status(app_handle)
         while  not status.is_terminal or elapsed > TIMEOUT
                time.sleep(INTERVAL)
                elapsed += INTERVAL
                status = self.runner.status(app_handle)
          
         if status.state == AppState.FAILED:
                for l in self.runner.log_lines(app_handle, streams=Stream.stderr):
                           print(l, file=sys.stderr)
                raise RuntimeError(status.structured_error_msg) 
         elif status.state == AppState.SUCCEEDED:
                   # run assertions about the end result/state of the application
                   # e.g. look for specific stdout or make sure the app produced a file (artifact) etc
                       
```

## Detailed Proposal
Make a test fixture so that the verbose test setup above can be succinctly expressed as:

```
class FooTest(ComponentTestFixture):
    
    def test_foo(self) -> None:
            # sprinkle some syntactic sugar so that the "run" looks similar to the CLI
            results = self.torchx.run(foo, args)

            # no need to check status; if the app fails, then we should print clear error messages, stderr, stack traces
            # of the app (e.g. deal with all the multiprocessing complexity)  to the console to give the user 
            # exact diagnostics information to debug their application
            
           self.assertEqual("something about stdout", results.stdout)
           # run other assertions about the app
```

## Alternatives
We could make the fixture less imperative as:

```
class MyComponentsTest(ComponentTestFixture):

[...]

PR #511, opened 2022-06-07
Test Helper to test components with a runner.

kurman:
Summary:
Exposing test helper method for covering component behavior using Runner and scheduler that is has a parity of running cli.

More details: https://github.com/pytorch/torchx/issues/370

Differential Revision: D36948958

### DIFF
--- a/torchx/components/component_test_base.py
+++ b/torchx/components/component_test_base.py
@@ -15,14 +15,32 @@ args.
 """
 
 import os
+import shutil
+import tempfile
+import time
 import unittest
 from types import ModuleType
+from typing import Any, Callable, Dict, Optional
+
+from torchx.runner import get_runner
+
+from torchx.specs import AppDef, AppStatus
 
 from torchx.specs.builders import _create_args_parser
 from torchx.specs.finder import get_component
 
 
 class ComponentTestCase(unittest.TestCase):
+    def setUp(self) -> None:
+        self.test_dir = tempfile.mkdtemp("torchx_component_test")
+
+        self.old_cwd = os.getcwd()
+        os.chdir(os.path.dirname(os.path.dirname(__file__)))
+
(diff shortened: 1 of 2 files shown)


torchx/components/component_test_base.py
```diff
@@ -15,14 +15,34 @@
...
+
+from torchx.specs import AppDef, AppStatus
 
 from torchx.specs.builders import _create_args_parser
 from torchx.specs.finder import get_component
 
 
 class ComponentTestCase(unittest.TestCase):
+    def setUp(self) -> None:
+        super().setUp()
```
d4l3k, line 35: don't need to call super for setUp/tearDown per the docs

torchx/components/component_test_base.py
```diff
@@ -15,14 +15,34 @@
...
+        self.test_dir = tempfile.mkdtemp("torchx_component_test")
```
d4l3k, line 36: I generally use TemporaryDirectory since it will clean up if the object gets destructed but in this case doesn't really matter either way

torchx/components/component_test_base.py
```diff
@@ -58,3 +78,43 @@ def validate(self, module: ModuleType, function_name: str) -> None:
         # this will raise an exception and the test will fail
         with self.assertRaises(SystemExit):
             _ = _create_args_parser(component_def.fn).parse_args(["--help"])
+
+    def run_component(
+        self,
+        component: Callable[..., AppDef],
+        args: Optional[Dict[str, Any]] = None,
```
d4l3k, line 85: wonder if it's slightly cleaner to use `**args` but that also mixes components vs run_component args so either is fine

d4l3k (approved):
looks good! -- I know you had discussed moving this to a new location. Did you change your mind?

PR #517, opened 2022-06-14
Replace partition by rpartition for parsing app_id

ljjsalt:
To parse a address like `"addr-of-cluster-app_id"`, the old code will fail since it takes `"addr"` as the address and `"of-cluster-app_id"` as the app id.
After replace `partition` by `rpartition`, it should work as intended.

Test plan:

Follow the original test.

### DIFF
--- a/torchx/schedulers/test/ray_scheduler_test.py
+++ b/torchx/schedulers/test/ray_scheduler_test.py
@@ -298,6 +298,17 @@ if has_ray():
                 job = req.request
                 self.assertEqual(job.requirements, reqs)
 
+        def test_parse_app_id(self) -> None:
+            test_addr_appid = [
+                ("0.0.0.0:1234-app_id", "0.0.0.1:1234", "app_id"), # (full address, address:port, app_id)
+                ("addr-of-cluster:1234-app-id", "addr-of-cluster:1234", "app-id"),
+                ("www.test.com:1234-app:id", "www.test.com:1234", "app:id"),
+            ]
+            for test_example, addr, app_id in test_addr_appid:
+                parsed_addr, parsed_appid = self._scheduler._parse_app_id(test_example)
+                self.assertEqual(parsed_addr, addr)
+                self.assertEqual(parsed_appid, app_id)
+
     class RayClusterSetup:
         _instance = None  # pyre-ignore[4]
 
--- a/torchx/schedulers/ray_scheduler.py
+++ b/torchx/schedulers/ray_scheduler.py
@@ -8,6 +8,7 @@ import dataclasses
 import json
 import logging
 import os
+import re
 import tempfile
 import time
 from dataclasses import dataclass, field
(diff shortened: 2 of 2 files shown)


d4l3k:
Hi, thanks for contributing!

The one thing I'm a bit worried about is in the cases where there's a dash in the ID since that's valid in all of our schedulers.

Since I believe the ray host address always contains the port (and IDs can't contain :) it might be best for us to do a combination where we look look for the first `:\d+-` and split on that instead.

Can pull out the partition logic into a helper method and would be nice to add some tests as well

```
foo-foo:1234-bar-bar
```

d4l3k (changes requested):
Hi, thanks for contributing!

The one thing I'm a bit worried about is in the cases where there's a dash in the ID since that's valid in all of our schedulers.

Since I believe the ray host address always contains the port (and IDs can't contain :) it might be best for us to do a combination where we look look for the first `:\d+-` and split on that instead.

Can pull out the partition logic into a helper method and would be nice to add some tests as well

```
foo-foo:1234-bar-bar
```

d4l3k:
Looks like that's the case--the ray integration tests are failing because of the right dashes 
```
ValueError: invalid literal for int() with base 10: '8265-torchx-utils-booth'
```

ljjsalt:
> Hi, thanks for contributing!
> 
> The one thing I'm a bit worried about is in the cases where there's a dash in the ID since that's valid in all of our schedulers.
> 
> Since I believe the ray host address always contains the port (and IDs can't contain :) it might be best for us to do a combination where we look look for the first `:\d+-` and split on that instead.
> 
> Can pull out the partition logic into a helper method and would be nice to add some tests as well
> 
> ```
> foo-foo:1234-bar-bar
> ```

Yes, this makes sense. So can we make sure that there is always at least one ":" in the host address and should we consider the situation that there are more than one ":"?

torchx/schedulers/test/ray_scheduler_test.py
```diff
@@ -298,6 +298,18 @@ def test_requirements(self) -> None:
...
+        def test_parse_app_id(self) -> None:
+            from itertools import product
+            addrs = ["0.0.0.0", "www.testexample.com", "addr-of-cluster"]
+            ports = ["65535", ]
+            app_ids = ["a-p-p_id", "app-i-d", "app_id", "app:id"]
+            for addr, port, app_id in product(addrs, ports, app_ids):
+                parsed_addr, parsed_app_id = self._scheduler._parse_app_id(
+                    f"{addr}:{port}-{app_id}"
+                )
+                self.assertEqual(parsed_addr, addr)
```
d4l3k, line 310: I think this is dropping the port -- port should still be included in the address

Would also be good to add an explicit test where there's no port specified and it should fall back to the old behavior for safety reasons

torchx/schedulers/test/ray_scheduler_test.py
```diff
@@ -298,6 +298,18 @@ def test_requirements(self) -> None:
                 job = req.request
                 self.assertEqual(job.requirements, reqs)
 
+        def test_parse_app_id(self) -> None:
+            from itertools import product
+            addrs = ["0.0.0.0", "www.testexample.com", "addr-of-cluster"]
+            ports = ["65535", ]
+            app_ids = ["a-p-p_id", "app-i-d", "app_id", "app:id"]
+            for addr, port, app_id in product(addrs, ports, app_ids):
```
d4l3k, line 306: product is clever but I think it's a bit easy to miss things since it's a bit hard to follow

no strong preference but might be better to just be very explicit and call _parse_app_id for each case even if it's a few more lines of code

d4l3k (changes requested):

torchx/schedulers/test/ray_scheduler_test.py
ljjsalt: My bad, I didn't see the instruction on how to run tests in the document.

torchx/schedulers/test/ray_scheduler_test.py
ljjsalt: OK, I have changed the test examples to explicit form.

d4l3k:
I made some fixups on this at https://github.com/pytorch/torchx/pull/523

@ljjsalt the commit should still link to your account if you add your @ibm.com email to your GitHub account

ljjsalt:
> 

ok no problem

d4l3k:
landed via https://github.com/pytorch/torchx/commit/5d1ad55bc52a0c0355bfc2719450882a54af6bf7

PR #519, opened 2022-06-15
docs: add airflow example

d4l3k:
This adds an example for how to use Airflow with TorchX. This example requires airflow server to be running on the local machine during the docs build process.

This also adds in a `AppStatus.raise_for_status()` method to match the requests/urllib3 behavior which just makes checking for whether or not a job failed easier.

Build is working locally -- might need to fiddle with CI to get airflow running there correctly still

Test plan:

```
cd docs && make html
```

Screenshot 2022-06-14 at 18-01-37 Airflow — PyTorch_TorchX main documentation

### DIFF
--- a/docs/source/pipelines/airflow.md
+++ b/docs/source/pipelines/airflow.md
@@ -1,0 +1,104 @@
+---
+jupyter:
+  jupytext:
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.3'
+      jupytext_version: 1.13.7
+  kernelspec:
+    display_name: Python 3
+    language: python
+    name: python3
+---
+
+# Airflow
+
+For pipelines that support Python based execution you can directly use the
+TorchX API. TorchX is designed to be easily integrated in to other applications
+via the programmatic API. No special Airflow integrations are needed.
+
+With TorchX, you can use Airflow for the pipeline orchestration and run your
+PyTorch application (i.e. distributed training) on a remote GPU cluster.
+
+```python
+import datetime
+import pendulum
+
+from airflow.utils.state import DagRunState, TaskInstanceState
+from airflow.utils.types import DagRunType
+from airflow.models.dag import DAG
+from airflow.decorators import task
+
+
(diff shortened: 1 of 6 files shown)


docs/source/pipelines/airflow.md
```diff
@@ -0,0 +1,97 @@
...
+    start_date=DATA_INTERVAL_END,
+    run_type=DagRunType.MANUAL,
+)
+ti = dagrun.get_task_instance(task_id="hello_torchx")
+ti.task = dag.get_task(task_id="hello_torchx")
+ti.run(ignore_ti_state=True)
+assert ti.state == TaskInstanceState.SUCCESS
+```
+
+If all goes well you should see `Hello, TorchX!` printed above.
```
kurman, line 97: What do you think on recommending links on things to do next after this example? For example apps that can be used here?

kurman (approved):
Really nice integration showcase.

docs/source/pipelines/airflow.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+To launch a TorchX job from Airflow you can create a Airflow Python task to
+import the runner, launch the job and wait for it to complete. If you're running
+on a remote cluster you may need to use the virtualenv task to install the
+`torchx` package.
+
+```python
+@task(task_id=f'hello_torchx')
+def run_torchx(echo):
```
priyaramani, line 45: Is the arg 'echo' used in the function? I see "foo" passed in from line 81 but the echo command used in the body at line 52 looks like a string literal

priyaramani (approved):

docs/source/pipelines/airflow.md
d4l3k: It is -- utils.sh runs a shell command which in this case is "echo"

docs/source/pipelines/airflow.md
d4l3k: done

docs/source/pipelines/airflow.md
d4l3k: sorry, you're right! I was thinking about the lower echo! my bad :)

PR #523, opened 2022-06-15
Replace partition by rpartition for parsing app_id

d4l3k:
Summary:
To parse a address like `"addr-of-cluster-app_id"`, the old code will fail since it takes `"addr"` as the address and `"of-cluster-app_id"` as the app id.
After replace `partition` by `rpartition`, it should work as intended.

Differential Revision: D37187555

Pulled By: d4l3k

### DIFF
--- a/torchx/schedulers/ray_scheduler.py
+++ b/torchx/schedulers/ray_scheduler.py
@@ -8,12 +8,13 @@ import dataclasses
 import json
 import logging
 import os
+import re
 import tempfile
 import time
 from dataclasses import dataclass, field
 from datetime import datetime
 from shutil import copy2, rmtree
-from typing import Any, cast, Dict, Iterable, List, Mapping, Optional, Set, Type  # noqa
+from typing import Any, cast, Dict, Iterable, List, Optional, Tuple  # noqa
 
 from torchx.schedulers.api import (
     AppDryRunInfo,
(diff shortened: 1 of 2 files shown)


torchx/schedulers/ray_scheduler.py
```diff
@@ -322,13 +323,25 @@ def wait_until_finish(self, app_id: str, timeout: int = 30) -> None:
...
+
             addr, _, app_id = app_id.partition("-")
+            return addr, app_id
+
+        def _cancel_existing(self, app_id: str) -> None:  # pragma: no cover
+            addr, app_id = self._parse_app_id(app_id)
             client = JobSubmissionClient(f"http://{addr}")
             client.stop_job(app_id)
 
         def _get_job_status(self, app_id: str) -> JobStatus:
```
kurman, line 343: It would be good to have explicit concepts that expresses app_id and  maybe_app_id_address.

kurman (approved):

torchx/schedulers/ray_scheduler.py
d4l3k: yeah, long term I'm not very happy with how we're adding the addr in here -- I think we might need to do some config refactors on this so there's a way to set a "scheduler" config for non submit methods. Most other schedulers pull context info like this from the environment variables but ray doesn't seem to do that

Issue #509, opened 2022-06-02
job launch hooks for linking to external services such as tensorboard


d4l3k:
## Description

When users are launching TorchX jobs they often want a way to provide links to view the results on external services. This commonly includes things like Tensorboard. We want to add a generic way to allow this from components.

### Specs

Update the AppDef to include an `external_links` field that can be macro filled.
https://github.com/pytorch/torchx/blob/main/torchx/specs/api.py#L341

```
external_links: Dict[str, str] = field(default_factory=dict)
```

We'll likely want to add a new "apply_app" macro method that can apply to a whole appdef to materialize these values.

https://github.com/pytorch/torchx/blob/main/torchx/specs/api.py#L146

### Usage

```py
def my_component() -> AppDef:
    return AppDef(
        name="train",
        roles=[
            Role(
               name="blah"
               args=[
                   "--tensorboard", f"s3://foo/bar/{macros.app_id}",
               ]
          ),
      ],
      external_links={
          "Tensorboard": f"https://tb_service/foo/bar/{macros.app_id}",
      }
    ),
```
        

### CLI

After the job launches the CLI should print the materialized links:

```
External Links:
  - Tensorboard: http://...
  - Latest checkpoint: s3://.../model.pt
```

### Envs

We may also want to add these links to the job metadata so it's easy for people viewing the job on the native scheduler to find and click the links.

## Alternatives

Runtime service/library to track experiment metadata. Hard since we don't currently have a runtime other than torchelastic and it still wouldn't show up on the user CLI

kiukchung:
Closing as not planned: no generic launch-hook mechanism is planned in `AppDef`. Trackers cover the linking case — register one with `@register.tracker()` and it receives the run — so a second hook surface for the same purpose is not worth the API.

PR #532, opened 2022-06-22
torchx: add metadata_urls to allow printing user links from the CLI

d4l3k:
Summary:
This adds a new `metadata_urls` to AppStatus and `metadata` to DescribeAppResponse. Metadata URLs are automatically returned from metadata values if they start with `http://` or `https://`. This isn't quite as clean as a dedicated URLs field that we discussed in https://github.com/pytorch/torchx/issues/509 but can work when calling status and not just during launch time.

This also enables a way for schedulers to return extra UI links and services that are scheduler specific but not available at launch time.

Differential Revision: D37353646


priyaramani (approved):

kiukchung:
The corresponding internal change (D37353646) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

Issue #503, opened 2022-05-25
add `torchx list` command and `Runner.list` APIs


d4l3k:
## Description

Add a `torchx list` and `Runner/Scheduler.list` methods. This would allow listing all jobs the user has launched and see their status when tracking multiple different jobs. 

## Motivation/Background

Currently users have to use the scheduler specific tools like `sacct/vcctl/ray job list` to see all of their jobs. Adding this would allow users to just interact via the torchx interface and not have to worry about interacting with other tools.

## Detailed Proposal

We'd likely want something similar to https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.list

Filters may be hard to support across all schedulers so we probably want to limit it to just a few common ones or none at all initially. We also want to filter so we only return torchx jobs instead of all jobs on the scheduler.

Limiting it to jobs that the user owns would also be nice to have though may not be feasible for all schedulers.

## Additional context/links

* https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.list
* https://slurm.schedmd.com/sacct.html
* https://docs.aws.amazon.com/batch/latest/APIReference/API_ListJobs.html
* https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#list_namespaced_custom_object
* https://docs.ray.io/en/master/cluster/jobs-package-ref.html#jobsubmissionclient

anishchopra:
+1 for this, would love to have this feature

d4l3k:
@anishchopra we'll likely add this in phases to support the various schedulers. What schedulers are you currently using just to prioritize and get early feedback?

We'd also be happy to get any other feedback you might have. Feel free to email me (tristanr@fb.com) or I can add you to the PT slack if you'd like to chat more

anishchopra:
> @anishchopra we'll likely add this in phases to support the various schedulers. What schedulers are you currently using just to prioritize and get early feedback?
> 
> We'd also be happy to get any other feedback you might have. Feel free to email me ([tristanr@fb.com](mailto:tristanr@fb.com)) or I can add you to the PT slack if you'd like to chat more

I'm currently testing out the local_docker and kubernetes schedulers. Definitely happy to provide more feedback and join that slack channel. FYI, I'm an engineer at Rivian, and currently exploring all the different ways we can launch distributed training jobs in the cloud.

priyaramani:
Hey @anishchopra 
[...]

PR #545, opened 2022-07-05
List API for local scheduler

priyaramani:
Add List API for local scheduler. This throws exception as apps launched by local scheduler are only known to the launching python process and are not persisted elsewhere. 
List API: https://github.com/pytorch/torchx/issues/503 

Test plan:
```pytest torchx/schedulers/local_scheduler_test.py```

### DIFF
--- a/test.py
+++ b/test.py
@@ -1,5 +1,0 @@
-from time import sleep
-
-print("hello")
-sleep(60)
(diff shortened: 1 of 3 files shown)


PaliC (approved):
lgtm

PR #546, opened 2022-07-07
List API - refactor out common logic

priyaramani:
Move common logic shared across schedulers(kubernetes, docker etc) like creating app_handle from app_id to runner list API from scheduler's list implementation 

Test plan:
Updated unit tests and ran pytest

### DIFF
--- a/torchx/runner/api.py
+++ b/torchx/runner/api.py
@@ -554,7 +554,11 @@ class Runner:
         """
         with log_event("list", scheduler):
             sched = self._scheduler(scheduler)
-            return sched.list()
+            app_handles = [
+                make_app_handle(scheduler, self._name, app_id)
+                for app_id in sched.list()
+            ]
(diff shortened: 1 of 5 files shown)


torchx/runner/api.py
```diff
@@ -554,7 +554,12 @@ def list(
         """
         with log_event("list", scheduler):
             sched = self._scheduler(scheduler)
-            return sched.list()
+            app_handles = []
+            app_ids = sched.list()
+            for app_id in app_ids:
```
kurman, line 559: We could use list comprehension to build `app_ids` for more idiomatic code:
`
app_handles = [make_app_handle(scheduler, self._name, app_id) for app_id in sched.list()]
`

priyaramani: Thanks for the suggestion. Updated!

kurman (approved):

Issue #548, opened 2022-07-09
Unix file modes not preserved for workspace files patched into docker images


yukunlin:
## 🐛 Bug

When files from a workspace is patched into a docker image, the unix file modes (i.e. `rwxr-xr-x` for an executable file) is not preserved.  Trying to execute the patched file leads to `Permission denied` errors because all patched files have mode `rw-r--r--`.

Module (check all that applies):
 * [ ] `torchx.spec`
 * [ ] `torchx.component`
 * [ ] `torchx.apps`
 * [ ] `torchx.runtime`
 * [ ] `torchx.cli`
 * [ ] `torchx.schedulers`
 * [ ] `torchx.pipelines`
 * [ ] `torchx.aws`
 * [ ] `torchx.examples`
 * [x] `torchx.workspace`

## To Reproduce

1. Create a new workspace:
   ```
   mkdir torchx_ws
   cd torchx_ws
   touch foo.sh
   ```
1. Make a simple bash script in `foo.sh`
   ```bash
   #!/bin/bash
   echo hi
   ```
1. Make `foo.sh` executable:
   ```
   chmod +x foo.sh
   ```
1. Patch and run `foo.sh` from workspace
   ```
   torchx run --scheduler local_docker utils.sh --image amazonlinux -- ./foo.sh
   ```
1. Check the permissions of `foo.sh` in the patched image
   ```
   torchx run --scheduler local_docker utils.sh --image amazonlinux -- ls -al ./foo.sh
   ```

## Expected behavior

`torchx run --scheduler local_docker utils.sh --image amazonlinux -- ./foo.sh` should have executed the script, but fails with a `Permission denied` error:

```
torchx 2022-07-08 21:10:57 INFO     Waiting for the app to finish...
torchx 2022-07-08 21:10:57 INFO     Job finished: FAILED
sh/0 sh: ./foo.sh: Permission denied
```

Checking the permissions with `torchx run --scheduler local_docker utils.sh --image amazonlinux -- ls -al ./foo.sh`, we see that the patched file isn't executable (despite the fact that the original file in the workspace is):

```
torchx 2022-07-08 21:12:12 INFO     Waiting for the app to finish...
torchx 2022-07-08 21:12:12 INFO     Job finished: SUCCEEDED
sh/0 -rw-r--r-- 1 root root 20 Jan  1  1970 ./foo.sh
```

## Environment

 - torchx version (e.g. 0.1.0rc1): torchx-0.3.0dev0 (from nightly build)
[...]

PR #549, opened 2022-07-09
Fixes pytorch/torchx#548

yukunlin:
Preserve Unix file mode when patching files into docker image.

Test plan:
* Tested with the repro steps in #548
* Verify that existing unit tests in `docker_workspace_test.py` still passes. Seems like the only way to write tests to check the mode is to create a [temporary directory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory). I'm not sure if the maintainers want tests that interact with the file system?

### DIFF
--- a/torchx/workspace/docker_workspace.py
+++ b/torchx/workspace/docker_workspace.py
@@ -7,6 +7,7 @@
 import io
 import logging
 import posixpath
+import stat
 import tarfile
 import tempfile
 from typing import Dict, IO, Mapping, Optional, Tuple, TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


d4l3k:
Thanks for sending the fix!

Feel free to create a temporary directory -- we do it in plenty of other areas. Always nice to use the in memory filesystem but a tempdir works just as well

yukunlin:
Added a commit with tests using temp directories.

d4l3k (approved):
thanks for contributing!

priyaramani (approved):

PR #559, opened 2022-07-14
ray: support expanding placement group on the fly, support torch elastic 

ljjsalt:
> This PR is for creating the placement group, fault tolerance should be supported in a different PR.

Support torch elastic launch on the ray cluster, and expand the placement group on the fly
1. `ray_driver.py`: first create a placement group that meets the minimum requirements, then create the rest of placement group(for 1 actor only) 1 by 1 until the maximum number of actors is reached. Instead of using multi-processing, I use the feature that the creation of a placement group is asynchronous can be monitored by `ray.wait`, and put the placement group creation together with other ray command actors in the same `ray.wait` to achieve placement group creation on the fly.
2. `dist.py`:  `--nnodes` argument now supports format like `--nnodes 1:4x2`, allow the number of nodes varies from 1 to 4.
3. some additional changes so `ray_driver.py` can retrieve the value from `--nnodes` arguments.

Something else need to discuss:
1. when the number of nodes are big, creating the placement group 1 by 1 is inefficient, since even the resources for all the placement groups are available, the script can started with minimum number of nodes before all the groups are created. A solution could be create all the placement group for the maximum requirements at the beginning.

Test plan:

1. Unit test for parsing the number of nodes arguments including the old and new grammars. (`test_ddp_parse_j`)
2. Utilize [ray cluster](https://docs.ray.io/en/latest/_modules/ray/cluster_utils.html) for testing the cluster, it was using CLI commands. (A reimplementation of class `RayClusterSetup`)
3. The testing of the behavior of TorchX on the ray cluster are including three parts:
    1. If placement groups can be created on the fly(detect the changing of the cluster). (`test_placement_group_creation`)
    2. If a command actor can be successfully created on create placement groups. (`test_command_actor_setup`)
    3. if a job can be launched on a command actor. ('test_ray_cluster')

### DIFF
--- a/torchx/components/dist.py
+++ b/torchx/components/dist.py
@@ -115,7 +115,7 @@ def ddp(
     # nnodes: number of nodes or minimum nodes for elastic launch
     # max_nnodes: maximum number of nodes for elastic launch
     # nproc_per_node: number of processes on each node
-    nnodes, max_nnodes, nproc_per_node, nnodes_rep = None, None, None, None
+    nnodes, max_nnodes, nproc_per_node, nnodes_rep = "", "", "", ""
     is_elastic = False  # elastic launch flag
     if re.match("\\d+:\\d+x\\d+", j):  # match 2:4x1
         nnodes_rep, nproc_per_node = j.split("x")
@@ -124,8 +124,8 @@ def ddp(
     elif re.match("\\d+x\\d+", j):  # match 2x1
         nnodes, nproc_per_node = j.split("x")
     elif re.match("\\d+", j):  # match 2
-        nnodes = 1
-        nproc_per_node = int(j)
+        nnodes = "1"
+        nproc_per_node = j
     else:
         raise ValueError(
             f"Invalid format for -j, usage example: 1:2x4 or 1x4 or 4. Given: {j}"
@@ -195,7 +195,7 @@ def ddp(
                 },
                 max_retries=max_retries,
                 mounts=specs.parse_mounts(mounts) if mounts else [],
-                nnodes_rep=nnodes_rep if is_elastic else None,
+                nnodes_rep=nnodes_rep,
             )
         ],
     )
--- a/torchx/components/dist.py
+++ b/torchx/components/dist.py
@@ -151,8 +151,11 @@ def ddp(
         rdzv_endpoint,
         "--rdzv_id",
         f"{macros.app_id}",
+
         "--nnodes",
-        str(nnodes),
+        # str(nnodes),
+        '1:2',
+
         "--nproc_per_node",
         str(nproc_per_node),
         "--tee",
--- a/torchx/components/dist.py
+++ b/torchx/components/dist.py
@@ -50,6 +50,7 @@ under the hood. Read more about torchelastic `here <https://pytorch.org/docs/sta
 Components APIs
 -----------------
 """
+import re
 import os
 import shlex
 from pathlib import Path
@@ -111,15 +112,24 @@ def ddp(
     if (script is None) == (m is None):
         raise ValueError("exactly one of --script and -m must be specified")
 
-    rep = j.split("x")
-    if len(rep) == 1:  # num replicas only
+    # nnodes: number of nodes or minimum nodes for elastic launch
+    # max_nnodes: maximum number of nodes for elastic launch
+    # nproc_per_node: number of processes on each node
+    nnodes, max_nnodes, nproc_per_node, nnodes_rep = None, None, None, None
+    is_elastic = False # elastic launch flag
+    if re.match('\d+:\d+x\d+', j): # match 2:4x1
+        print("grats, it's elastic launch")
+        nnodes_rep, nproc_per_node = j.split('x')
+        nnodes, max_nnodes = nnodes_rep.split(':')
+        is_elastic = True
+    elif re.match('\d+x\d+', j): # match 2x1
+        print("normal launch")
+        nnodes, nproc_per_node = j.split('x')
+    elif re.match('\d+', j): # match 2
         nnodes = 1
-        nproc_per_node = int(rep[0])
-    elif len(rep) == 2:
-        nnodes = int(rep[0])
-        nproc_per_node = int(rep[1])
+        nproc_per_node = int(j)
     else:
-        raise ValueError(f"Invalid format for -j, usage example: 1x4. Given: {j}")
+        raise ValueError(f"Invalid format for -j, usage example: 1:2x4 or 1x4 or 4. Given: {j}")
 
     if script:
         # script name/module no extension
@@ -130,9 +140,11 @@ def ddp(
         raise ValueError("failed to compute role_name")
 
     rdzv_backend = "c10d"
-    if nnodes == 1:
+    if (not is_elastic and int(nnodes) == 1) or (is_elastic and int(nnodes)==0 and int(max_nnodes) == 1):
         # using port 0 makes elastic chose a free random port which is ok
         # for single-node jobs since all workers run under a single agent
+        # When nnodes is 0 and max_nnodes is 1, it's stil a single node job
+        # but pending until the resources become available
         rdzv_endpoint = "localhost:0"
     else:
         rdzv_endpoint = _noquote(f"$${macros.rank0_env}:{rdzv_port}")
@@ -151,11 +163,8 @@ def ddp(
         rdzv_endpoint,
         "--rdzv_id",
         f"{macros.app_id}",
-
         "--nnodes",
-        # str(nnodes),
-        '1:2',
-
+        nnodes_rep if is_elastic else nnodes, # Use nnodes_rep only when elastic launch
         "--nproc_per_node",
         str(nproc_per_node),
         "--tee",
@@ -175,7 +184,7 @@ def ddp(
                 name=role_name,
                 image=image,
                 entrypoint="bash",
-                num_replicas=nnodes,
+                num_replicas=int(max_nnodes) if is_elastic else int(nnodes),
                 resource=specs.resource(cpu=cpu, gpu=gpu, memMB=memMB, h=h),
                 args=["-c", _args_join(cmd)],
                 env=env,
@@ -184,6 +193,7 @@ def ddp(
                 },
                 max_retries=max_retries,
                 mounts=specs.parse_mounts(mounts) if mounts else [],
+                # nnodes_rep=nnodes_rep if is_elastic else None,
             )
         ],
     )
--- a/torchx/components/dist.py
+++ b/torchx/components/dist.py
@@ -50,8 +50,8 @@ under the hood. Read more about torchelastic `here <https://pytorch.org/docs/sta
 Components APIs
 -----------------
 """
-import re
 import os
+import re
 import shlex
 from pathlib import Path
 from typing import Dict, Iterable, List, Optional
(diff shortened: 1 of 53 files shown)


torchx/components/dist.py
```diff
@@ -128,15 +129,24 @@ def ddp(
...
         raise ValueError("exactly one of --script and -m must be specified")
 
-    rep = j.split("x")
-    if len(rep) == 1:  # num replicas only
-        nnodes = 1
-        nproc_per_node = int(rep[0])
-    elif len(rep) == 2:
-        nnodes = int(rep[0])
-        nproc_per_node = int(rep[1])
+    # nnodes: number of nodes or minimum nodes for elastic launch
```
d4l3k, line 132: can we pull this logic out into a helper method? that way we'll be able to test it directly

```
min_replicas, max_replicas, nproc_per_node = parse_nnodes(...)
```

torchx/components/dist.py
```diff
@@ -147,9 +157,13 @@ def ddp(
         raise ValueError("failed to compute role_name")
 
     rdzv_backend = "c10d"
-    if nnodes == 1:
+    if (not is_elastic and int(nnodes) == 1) or (
+        is_elastic and int(nnodes) == 0 and int(max_nnodes) == 1
```
d4l3k, line 161: I don't think we need to support the zero node case. If you just ask for 1 node it'll be pending until resources are available regardless

torchx/specs/api.py
```diff
@@ -334,6 +334,7 @@ class Role:
     mounts: List[Union[BindMount, VolumeMount, DeviceMount]] = field(
         default_factory=list
     )
+    nnodes_rep: str = ""  # the representation of nnodes for elastic launch
```
d4l3k, line 337: can we change this to be `min_replicas: Optional[int] = None` and thus you'd use f"{min_replicas}:{num_replicas}" as the range for elastic? I think that's cleaner

Also document it

torchx/components/dist.py
```diff
@@ -172,7 +186,7 @@ def ddp(
         "--rdzv_id",
         f"{macros.app_id}",
         "--nnodes",
-        str(nnodes),
+        nnodes_rep if is_elastic else nnodes,  # Use nnodes_rep only when elastic launch
```
d4l3k, line 189: can we get rid of is_elastic? If min nodes == max nodes setting `f"5:5"` seems like it's fine

torchx/schedulers/ray/ray_driver.py
```diff
@@ -127,17 +137,33 @@ def create_command_actors(
 
 
 def main() -> None:  # pragma: no cover
+
     actors: List[RayActor] = load_actor_json("actors.json")
+    if re.match("\\d+:\\d+", actors[0].nnodes_rep):
```
d4l3k, line 142: outside of the component file we should avoid regular expressions (use a typed min_replicas instead)

torchx/schedulers/ray/ray_driver.py
```diff
@@ -127,17 +137,33 @@ def create_command_actors(
...
+    command_actors: List[CommandActor] = create_command_actors(actors[:MIN_NNODES], pg)
+    created += MIN_NNODES
 
     active_workers = [
         command_actor.exec_module.remote()  # pyre-ignore
         for command_actor in command_actors
     ]
 
+    # keep creating placement groups until the maximum number of actors is reached
+    if created < MAX_NNODES:
```
d4l3k, line 162: is this supposed to be a while loop or if?

we also don't increment created here though I'm wonder if it's better to delete this block and fold it into the below loop since we also have logic for it there

torchx/schedulers/ray/ray_driver.py
```diff
@@ -147,7 +173,23 @@ def main() -> None:  # pragma: no cover
         # If a failure occurs the ObjectRef will be marked as completed.
         # Calling ray.get will expose the failure as a RayActorError.
         for object_ref in completed_workers:
-            ray.get(object_ref)
+            result = ray.get(object_ref)
+            if isinstance(result, PlacementGroup):
```
d4l3k, line 177: can you explain how this works? not super familiar with the API but seems like we shouldn't be creating workers after some replicas have already finished?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -7,6 +7,7 @@
 import json
```
d4l3k, line 7: would be good to add a longish docstring explaining the elastic scheduling algorithm and all of the edge cases and how they're handled

d4l3k (changes requested):
left some general comments/questions about behaviors -- I didn't nit too much just looking at the high level right now

torchx/components/dist.py
ljjsalt: Done

torchx/components/dist.py
ljjsalt: Done

torchx/specs/api.py
ljjsalt: Done

torchx/components/dist.py
ljjsalt: Done

torchx/schedulers/ray/ray_driver.py
ljjsalt: Done

torchx/schedulers/ray/ray_driver.py
ljjsalt: 1. Once you call a remote function like `obj = fn.remote()`, the return value `obj` is just an object reference, ithe function `fn` may haven't been executed yet. You get the value of this reference only if you call `ray.get()`, and block until the function has been completed.
2. `ray.util.placement_group(bundle)`, bundle is a list of dict include the resources needed for this placement group. And this call is non-blocking, you get the reference of the placement group directly without waiting. The placement group can be created or still pending, but they all have a reference. `obj = pg.ready()` is actually a *remote call* to test if the placement group has been created, and it's also an object reference, you have to use `ray.get(obj)` to get the return value(`Optional[PlacementGroup]`, where `PlacemenGroup` is the reference to the placement group just mentioned) of `pg.ready()`.
3. `actor.exec_module.remote()` is a remote call that let the command actor run the script the user provided. It doesn't return any value, but you still have an object reference for `None`. (Every `ray.remote` call) returns an object reference, you only get the value once you use `ray.get`)
4. Since both `actor.exec_module.remote()` and `pg.ready()`  return object references, we can use `ray.wait()` to wait on both of them. `ray.wait` is blocking and partial waiting, as long as one of the function has completed, it returns the completed task and the tasks are still waiting.
5. The logic in the loop is we think both placement group creation and command actor execution as a task, based on the type of its return value, we know if it's an actor completed or a new placement group is created. If it's a placement group created, we can put a command actor on it.

torchx/specs/api.py
d4l3k: this seems like it hasn't changed?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -130,24 +139,43 @@ def main() -> None:  # pragma: no cover
...
     ray.init(address="auto", namespace="torchx-ray")
-    pg: PlacementGroup = create_placement_group(actors)
-    command_actors: List[CommandActor] = create_command_actors(actors, pg)
 
-    active_workers = [
-        command_actor.exec_module.remote()  # pyre-ignore
-        for command_actor in command_actors
-    ]
+    pending_placement_groups = [
+        create_placement_group(actors[i : i + 1]) for i in range(len(actors))
```
d4l3k, line 144: this should request min replicas first right instead of a one size pg for each replica?

torchx/schedulers/ray/ray_driver.py
d4l3k: Thanks! I missed the pg.ready() bit, sounds reasonable. The docstring doesn't need to include quite this much detail just the high level

torchx/schedulers/ray/ray_driver.py
```diff
@@ -130,24 +139,43 @@ def main() -> None:  # pragma: no cover
...
-            ray.get(object_ref)
+        for object_ref in completed_tasks:
+            # completed tasks contains two kinds of tasks:
+            # 1) placement group creation; 2) command actor execution
+            result = ray.get(object_ref)  # pyre-ignore
+            if isinstance(result, PlacementGroup):
+                new_actor: CommandActor = create_command_actors(
+                    actors[
+                        0:1
+                    ],  # use actors[0:1] since every actor in the list is the same
```
d4l3k, line 170: this isn't guaranteed for heterogeneous jobs with multiple roles

torchx/schedulers/ray/ray_driver.py
atinsood: what @d4l3k said, but also I think you have created this method `create_placement_group_async` but are never calling it. your code is still calling the `create_placement_group` in the main function

atinsood: since we are inrementally creating the placement group, do we/should we even deal with the fact that if a pg creation fails we just get cancel the creation of the rest of the pgs, or do we just take that as a separate task

ljjsalt: It's old code, in the latest code, it's async.

torchx/schedulers/ray/ray_driver.py
atinsood: I am also still a little reluctant on the fact that both the pg tasks and actor tasks are going into the same list and we are using instance type to separate out what logic we should be doing.

`completed_tasks, active_tasks = ray.wait(active_tasks)` now means that both pg and the actor will wait for the same period of time. 

there also doesn't seem be any exit out if the pg creation never happens, the loop will keep infinitely running even if the training has been completed. 

I like the idea of creating pg via tasks, but I think trying to accomplish everything in the same loop os introducing too many complications and we be better off using asyncio and treat these as 2 separate loops running in parallel

torchx/schedulers/ray/ray_driver.py
atinsood: is the new code on a different branch? should it be in this PR os a new PR. 

also other alternative might be the take these 2 loops and put them in two different actors that are basically running these loops in parallel.

torchx/specs/api.py
ljjsalt: It's actually removed now. It was used to let the `ray_driver.py` know the min:max numbers of nodes, but now it's not necessary.

torchx/schedulers/ray/ray_driver.py
ljjsalt: @atinsood Actually, those tasks that from the command actors will never be finished until the entire training process is finished. The command actors' tasks will always return a None, while the placement groups will always return a reference to the placement group.
```
            result = ray.get(object_ref)  # pyre-ignore
            if isinstance(result, PlacementGroup):
                new_actor: CommandActor = create_command_actors(
                    [
                        actors[
                            pg_ids[result.id]
                        ],  # find the actor of a placement group based on pg_id
                    ],  # must be a list
                    result,
                )[0]
                active_tasks.append(
                    new_actor.exec_module.remote()  # pyre-ignore
                )  # add a command actor execution task
            else:
                # if the result is None, it's returned by a completed worker, and the job is completed
                return
```
There is an `else` branch, which means if the result is None(non-placement group). it is returned by a command actor, and it means the entire training process is finished, then the return will end this process.

ray.wait on the command actors doesn't do anything but wait the actors to finish their jobs. Just like waiting for the child process to be completed.

torchx/schedulers/ray/ray_driver.py
ljjsalt: done

torchx/schedulers/ray/ray_driver.py
ljjsalt: done, now it's using a dictionary which uses PlacementGroup Id as the key and actor index as the value to match the placement groups and the actors.

PR #561, opened 2022-07-18
List API for Ray scheduler 

priyaramani:
Implement list API for Ray shceduler

Test plan:
Unit and integration test
```
(venv38) ubuntu@ip-172-31-24-88:~/rsync/torchx$ torchx list -s ray
ray://torchx/127.0.0.1:8265-123
```

### DIFF
--- a/torchx/schedulers/ray_scheduler.py
+++ b/torchx/schedulers/ray_scheduler.py
@@ -397,7 +397,17 @@ if _has_ray:
             return iterator
 
         def list(self) -> List[str]:
-            raise NotImplementedError()
+            address = os.getenv("RAY_ADDRESS")
+            if not address:
+                raise Exception(
+                    "RAY_ADDRESS env variable is expected to be set to list jobs on ray scheduler."
+                    " See https://docs.ray.io/en/latest/cluster/jobs-package-ref.html#job-submission-sdk for more info"
+                )
+            client = JobSubmissionClient(address)
+            jobs_dict = client.list_jobs()
+            ip = address.split("http://", 1)[-1]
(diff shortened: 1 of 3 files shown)


torchx/schedulers/ray_scheduler.py
```diff
@@ -397,7 +397,16 @@ def log_iter(
             return iterator
 
         def list(self) -> List[str]:
-            raise NotImplementedError()
+            address = os.getenv("RAY_ADDRESS")
+            if address is None:
+                raise Exception(
+                    "RAY_ADDRESS env variable is expected to be set to list jobs on ray scheduler"
```
kurman, line 403: What about just keeping "RAY_ADDRESS env variable is missing" message?

priyaramani: May be, but if we are also going to throw for empty value for RAY_ADDRESS, I'm thinking of a message like "RAY_ADDRESS env variable is expected to be set validly"

torchx/schedulers/ray_scheduler.py
```diff
@@ -397,7 +397,17 @@ def log_iter(
             return iterator
 
         def list(self) -> List[str]:
-            raise NotImplementedError()
+            address = os.getenv("RAY_ADDRESS")
+            if not address:
+                raise Exception(
+                    "RAY_ADDRESS env variable is expected to be set validly to list jobs on ray scheduler."
```
kurman, line 403: "validly" reads weird, what about:
expecting valid RAY_ADDRESS env variable to be set to list jobs on ray scheduler.

PR #564, opened 2022-07-21
Use raw content for kubectl apply

JarnoRFB:
Otherwise, the error `error: error parsing https://github.com/pytorch/torchx/blob/main/resources/etcd.yaml: error converting YAML to JSON: yaml: line 28: mapping values are not allowed in this context` is raised.


d4l3k:
@JarnoRFB thanks for contributing!

As a heads up -- that line is actually outdated, you don't need to install etcd to have the builtin `dist.ddp` component work out of the box. Just volcano.

Let me know if you run into any other issues!

PR #572, opened 2022-08-02
[Ray] Add elasticity and fault tolerance features to jobs launched on ray cluster

ljjsalt:
Two features for Elastic Distributed Training are added to job launched by TorchX on Ray Cluster in this PR:
1. Fault Tolerance - Node failure throws RayActorError which can be captured. Placement Groups have built-in fault tolerance, and can recover from node failure automatically.
2. Elasticity - the execution of placement groups are pending tasks that will be scheduled by GCS when resources become availiable.

The logic of the new `ray_driver.py` is in the plot below:

Drawing 2022-08-02 15 51 39 excalidraw

Test plan:

[Note]: This PR is related to the previous PR #559. All future changes will be submitted to this PR.


torchx/schedulers/ray/ray_driver.py
```diff
@@ -77,77 +111,184 @@ def load_actor_json(filename: str) -> List[RayActor]:
...
+        self.rank_0_port: Optional[int] = None
+        self.min_nnodes, self.max_nnodes = parse_nnodes_rep(replicas)  # pyre-ignore
+
+        self.placement_groups: List[PlacementGroup] = []
+        self.actor_info_of_id: Dict[str, ActorInfo] = {}
+        self.active_tasks: List["ray.ObjectRef"] = []
+
+    def init_placement_groups(self) -> None:
+        """Initialize all placement groups needed for this job"""
+        replica_ix_of_pg: List[int] = [0] + list(
```
d4l3k, line 145: it might be better to explicitly make the calls here. + we need to wait for the first placement group before creating the remainders otherwise the smaller groups might get scheduled first (unless ray does fifo queuing, which if it does we should document that)

```
initial = create_placement_group_async(self.replicas[0: self.min_nodes])
pg.wait(itimeout_seconds=...)
self.groups = []
for range(self.min_nodes, self.max_nodes):
    self.groups.append(create_placement_group_async(self.replicas[i:i+1]))
```

torchx/schedulers/ray/ray_driver.py
```diff
@@ -77,77 +111,184 @@ def load_actor_json(filename: str) -> List[RayActor]:
...
+                            )
+                        command_actors_count += 1
+                    elif isinstance(result, TaskCompleted):
+                        need_more_actors = False  # don't need more actors
+                        command_actors_count -= 1  # 1 completed command actor
+                        self.pop_actor_info(result.id)
+                        if (
+                            command_actors_count == 0
+                        ):  # all the command actors have finished
+                            break  # exit
```
d4l3k, line 235: this only breaks the inner loop -- is there a case where active_tasks >0 but command_actors_count == 0?

torchx/schedulers/ray/ray_driver.py
```diff
@@ -77,77 +111,184 @@ def load_actor_json(filename: str) -> List[RayActor]:
...
+            )
+            for i in range(len(replica_ix_of_pg) - 1)
+        ]
+
+    def pop_actor_info(self, actor_id: str) -> ActorInfo:
+        """Remove the info of  (dead) command actor"""
+        return self.actor_info_of_id.pop(actor_id)
 
+    def reschedule_actor(self, actor_id: str) -> None:
+        """Rescheule a failed actor"""
```
d4l3k, line 161: we should check the role max_retries/retry_policy here and throw an error if any of the workers exits more than N times

torchx/schedulers/test/ray_scheduler_test.py
```diff
@@ -322,44 +325,81 @@ def test_list_throws_without_address(self) -> None:
                 self._scheduler.list()
```
d4l3k, line 325: would be good to add some "state machine" mock tests that runs through that driver loop step by step

torchx/schedulers/ray/ray_driver.py
```diff
@@ -77,77 +111,184 @@ def load_actor_json(filename: str) -> List[RayActor]:
...
+            pg = self.placement_groups[pg_ix]  # find the created placement group
+            replica = self.replicas[i]
+            self.create_and_schedule_actor(pg, replica)
+
+    def run(self) -> None:
+        result: Optional[PlacementGroup]  # execution result
+        need_more_actors: bool = True  # if need more actors
+        command_actors_count: int = 0  # number of created command actors
+        # Await return result of remote ray function and initialize new command actors
+        while len(self.active_tasks) > 0:
```
d4l3k, line 200: can we split loop out from run so it has a "_step()" and "run()" method so we can manually step through the behavior from a mocked unit test?

d4l3k (changes requested):
getting into a nicer shape :) I like how you used different return values as state machine inputs

torchx/schedulers/ray/ray_driver.py
```diff
@@ -77,77 +111,184 @@ def load_actor_json(filename: str) -> List[RayActor]:
...
+        return len(actors), len(actors)
+    if ":" in rep:
+        min_nnodes, max_nnodes = rep.split(":")
+        min_nnodes, max_nnodes = int(min_nnodes), int(max_nnodes)
+    else:
+        min_nnodes, max_nnodes = int(rep), int(rep)
+    return min_nnodes, max_nnodes
+
+
+def parse_actor_id_from_error(err: RayActorError) -> str:
```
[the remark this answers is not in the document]
ljjsalt: It seems actor_id is the only thing we get from the exception without changing ray's code. Only error messages. https://github.com/ray-project/ray/blob/4c5c5763efff50bf9f76ae73a8c0073183dbb0cc/python/ray/exceptions.py#L233

torchx/schedulers/ray/ray_driver.py
```diff
@@ -77,77 +111,184 @@ def load_actor_json(filename: str) -> List[RayActor]:
...
+                            break  # exit
+                    else:
+                        raise RuntimeError(
+                            "Ray actor returns unkown type. This is most likely bug in torchx"
+                            "Open issue at https://github.com/pytorch/torchx"
+                        )
+                except RayActorError as err:
+                    # reschedule the failed command actor (node failure)
+                    command_actors_count -= 1  # remove the failed actor
+                    failed_actor_id: str = parse_actor_id_from_error(err)
```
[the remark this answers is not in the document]
ljjsalt: I checked, but I think there isn't actor related information. https://github.com/ray-project/ray/blob/4c5c5763efff50bf9f76ae73a8c0073183dbb0cc/python/ray/includes/object_ref.pxi#L38

But we can create some issues about this one and the exception one.

torchx/schedulers/ray/ray_driver.py
ljjsalt: Yes, active_tasks may contain some actors which haven't been scheduled successfully in a placement group, even after the job has finished. I guess I have to use return here.

torchx/schedulers/ray/ray_driver.py
ljjsalt: Ray doesn't do FIFO, will change.

torchx/schedulers/ray/ray_driver.py
ljjsalt: Actually, here if we add a max_retries, it actually means how many node failures we can endure, so it's different from the max_retries in the ray actor's context.

ljjsalt: As long as the number of working nodes is bigger than min_nnodes, we should allow infinite node failures, unless nodes are failing too frequently. That can bea retry_policy.

atinsood:
@ljjsalt @d4l3k 

I have been thinking more about this and I am wondering if ray.util.queue is a better way of implementing this. 

you basically create 2 actors, PlacementGroupManager actor and CommandManager actor and exchange information between them using ray queue.

PGManager group actor is responsible for creating the placement group and then putting a message in the queue for the CommandManager actor to process which can then create the command actor. this helps us keep the knowledge of pg creation and command actor creation compartmentalized in these actors. 

additionally, we can @remote these fns so both of these can be run in parallel

https://docs.ray.io/en/releases-1.2.0/advanced.html#message-passing-using-ray-queue

ljjsalt:
> 
The reason that we should use Placement Group is: if we only use ray actors here, even though we can rerun it when it throws a RayActorError. But this job could potentially lose the computation resource it used to have.

For example, there are 3 nodes each with 1 cpu,  job 1 requires minimum nodes of 3, and it’s running with all 3 nodes, another job 2 launched later and it requires 1 cpu, and becomes a pending task. Once a node failure happens, we have to rerun the actor which becomes a pending task after job 2, and job 2 will take that node, job 1 will fail since there aren't enough nodes to restart.

I didn't get it why to use queue, since we didn't use any inter process communication here.

atinsood:
```
@ray.remote
class PlacementGroupManager(object):
    def __init__(self, min_nodes, max_nodes, queue):
        self.queue = queue
        pass

    def run() -> PlacementGroup:
        # step 1. create a PG with min replicas

        #put the initial pg in the queue
        self.queue.put(ready)

        # step 2. while loop to keep create the rest of PGs incrementatlly,

        #go through the rest of the pgs one by one and keep adding them to the queue
        #we can deal with the logic of pg failure here or when to stop pg creation if needed


@ray.remote
class CommandActorManager(object):
    def __init__(self, queue, active_workers):
        self.queue = queue
        self.active_workers = []
        pass

    def run(self):
        #some logic on when to stop the while loop, either a poison pill or a signal actor
        while True:
            self.queue.get() #get notified that the pg was created
            # logic to create the command actor and goes here
            active_workers.append(command_actor.exec_module.remote()) # or may be use another queue

            #once you are done reading all the PGs from the queue or if the queue 





def main() -> None:  # pragma: no cover
    actors: List[RayActor] = load_actor_json("actors.json")
    # pyre-fixme[16]: Module `worker` has no attribute `init`.
    ray.init(address="auto", namespace="torchx-ray")
    q = Queue() # we can set the set of the queue upfront
    pg_manager = PlacementGroupManager(2,4,q) #min, max and queue
    pg_manager.run.remote()
    cmd_actor_manager = CommandActorManager(q, active_workers)
    cmd_actor_manager.run.remote()
    # Await return result of remote ray function
    while len(active_workers) > 0:
        _logger.info(f"running ray.wait on {active_workers}")

        # pyre-fixme[16]: Module `worker` has no attribute `wait`.
        completed_workers, active_workers = ray.wait(active_workers)
        # If a failure occurs the ObjectRef will be marked as completed.
        # Calling ray.get will expose the failure as a RayActorError.
        for object_ref in completed_workers:
            ray.get(object_ref)


```

I was thinking something like this, this is not the correct code, but more of a thought process on how to do interaction between the logic that is creating placement groups and the logic that is creating command actors and how to keep them compartmentalized

ljjsalt:
> @ljjsalt @d4l3k
> 
> I have been thinking more about this and I am wondering if ray.util.queue is a better way of implementing this.
> 
> you basically create 2 actors, PlacementGroupManager actor and CommandManager actor and exchange information between them using ray queue.
> 
> PGManager group actor is responsible for creating the placement group and then putting a message in the queue for the CommandManager actor to process which can then create the command actor. this helps us keep the knowledge of pg creation and command actor creation compartmentalized in these actors.
> 
> additionally, we can @Remote these fns so both of these can be run in parallel
> 
> https://docs.ray.io/en/releases-1.2.0/advanced.html#message-passing-using-ray-queue



> 
```python
        # step 2. while loop to keep create the rest of PGs incrementatlly,

        #go through the rest of the pgs one by one and keep adding them to the queue
        #we can deal with the logic of pg failure here or when to stop pg creation if needed
```

Actually, this step is not necessary. First, increasing PG one by one can be a problem once the number of nodes is large.
Screen Shot 2022-08-03 at 12 31 14 AM
Please read the description of the X axis, it only makes the time that a PG creation event added to GCS pending queue longer.

d4l3k:
If we did want to separate the concerns here with scheduling vs retries we could use the builtin task max_retries config though that has some other implications. 

I'm not sure that using a queue with two actors would simplify this logic -- state machines that you can step through are nice from a testing perspective

ljjsalt:
> If we did want to separate the concerns here with scheduling vs retries we could use the builtin task max_retries config though that has some other implications.
> 
> I'm not sure that using a queue with two actors would simplify this logic -- state machines that you can step through are nice from a testing perspective

But we are just running schedule(which cannot go wrong) and `exec_module` functions(which cannot go wrong unless script fails), it it necessary to use max_retries here? Of course, the feature can be easily added.

PR #592, opened 2022-08-30
Bump torchx pyre-check-nightly version

stroxler:
Summary: Bump pyre-nightly to the latest release, to prevent errors due to error suppressions.

Differential Revision: D39139755


stroxler:
It turns out we had a problem with a bad binary build (incorrect `ld` path).

Bumping again, to the version that worked to fix pyre-check CI

PR #627, opened 2022-10-26
[bugfix] add 'tabulate' to the requirements.txt since the CLI cmd_list uses it

kiukchung:
Fixes #{issue number}


d4l3k (approved):
thanks for the fix!

d4l3k:
@kiukchung can you sign the new CLA?

PR #632, opened 2022-10-27
(torchx/workspace) Improve docker push logging using bash escape sequence (mimicks docker CLI) to prevent log spamming when pushing for the first time

kiukchung:
See screenshot for before and after

Before:
image

After:
image

image


kiukchung:
@kurman , @priyaramani , @d4l3k - can one of you take a look?

kurman (approved):
Nice, looks much better.

torchx/workspace/docker_workspace.py
```diff
@@ -166,13 +167,42 @@ def push_images(self, images_to_push: Dict[str, Tuple[str, str]]) -> None:
             log.info(f"pushing image {repo}:{tag}...")
             img = client.images.get(local)
             img.tag(repo, tag=tag)
-            for line in client.images.push(repo, tag=tag, stream=True, decode=True):
+            ids = []
+            for event in client.images.push(repo, tag=tag, stream=True, decode=True):
```
d4l3k, line 171: can we pull this printing logic out into a helper method that takes the iterator from push?

would also be nice to have a unit test

d4l3k (changes requested):
Thanks for contributing! This looks really nice to have just needs a couple of small changes :)

torchx/workspace/docker_workspace.py
kiukchung: Done. Pulled the print logic out to a function called `print_push_event()` and added unittest for it with a sample event file.

d4l3k (approved):
thanks for the test!

PR #638, opened 2022-10-28
(torchx/resources) Add g4d, g5, p4d, p3 (rest of them), and trn1 types

kiukchung:
Adding complete instance types for g4d, g5, p4d, p3 and trn1 (trainium)

Test plan:
Added unittests


d4l3k (approved):
LGTM

Issue #636, opened 2022-10-28
[torchx/scheduler - aws_batch] list only returns 100 jobs


kiukchung:
## 🐛 Bug

The aws_batch_scheduler's `list` API only returns up to 100 jobs in the job queue because it does not properly use the paginator to list ALL jobs in the job queue. This happens because the batch scheduler in torchx does not properly make use of the pagination token that applies to most `list` and `describe` API calls in boto3.

See: https://github.com/pytorch/torchx/blob/main/torchx/schedulers/aws_batch_scheduler.py#L599

FWIW if there were more than 100 job queues then you might get unlucky and either: 

1. not find the jobqueue specified in the runopts since the [`describe_job_queues()`](https://github.com/pytorch/torchx/blob/main/torchx/schedulers/aws_batch_scheduler.py#L580) call also does not paginate correctly. 
2. -- or --  miss a bunch of jobs in other job queues (currently the behavior in torchx is that jobs across ALL job queues are listed - this might not be the best behavior).

> However since batch has a service limit of 50 JQs per account (unless the customer asks for a limit increase) we got lucky that this condition is never hit. 

The fix is quite simple: use boto3 [paginators](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/paginators.html) for any API calls that return paginated results (typically for `list` and `describe` calls on resources). In the case of batch, you would do:

```
for list_job_resp in boto3.client("batch").get_paginator("list_jobs"):
   # do what we were doing originally...
   pass
```

Module (check all that applies):
 * [ ] `torchx.spec`
 * [ ] `torchx.component`
 * [ ] `torchx.apps`
 * [ ] `torchx.runtime`
 * [ ] `torchx.cli`
 * [x] `torchx.schedulers`
 * [ ] `torchx.pipelines`
 * [ ] `torchx.aws`
 * [ ] `torchx.examples`
 * [ ] `other`

## To Reproduce

1. Get more than 100 jobs running in the batch job queue
1. Try listing it with `torchx list -s aws_batch`
1. Only 100 will be listed

## Expected behavior

All jobs in the job queue are returned or a proper pagination token is returned so that subsequent calls can pass the pagination token to iterate through all the jobs in the job queue.

## Environment

N/A

## Additional context

1. boto3 batch docs: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch.html

PR #642, opened 2022-11-01
(torchx/aws_batch) Properly describe-JQs and list-jobs using boto3 pa…

kiukchung:
Fixes https://github.com/pytorch/torchx/issues/636

Test plan:

Added/modified unittests where appropriate

```
pytest

================================================================================== 492 passed, 52 warnings in 120.35s (0:02:00) ==================================================================================
```

### DIFF
--- a/torchx/schedulers/aws_batch_scheduler.py
+++ b/torchx/schedulers/aws_batch_scheduler.py
@@ -473,13 +473,14 @@ class AWSBatchScheduler(DockerWorkspaceMixin, Scheduler[AWSBatchOpts]):
     def _get_job_id(self, app_id: str) -> Optional[str]:
         queue, name = app_id.split(":")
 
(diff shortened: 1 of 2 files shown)


kiukchung:
cc) @priyaramani  can you take a look?

d4l3k (approved):
@kiukchung if you hit performance issues might be good to change list to return an iterable so we can stream the values out but I don't see an issue with this code as written

kiukchung:
> @kiukchung if you hit performance issues might be good to change list to return an iterable so we can stream the values out but I don't see an issue with this code as written

I regularly see anywhere between 150 - 500 jobs at any given point in time so not hitting any perf issues since the list would have less than 500 elements. Will keep an eye out for any perf issues as I'm using this, thanks.

PR #664, opened 2022-12-08
Update to newer Accuracy API

priyaramani:
Instead of pinning torchmetrics library, use their updated latest APIs

Test plan:
CI


kiukchung:
The corresponding internal change (D41882948) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #667, opened 2022-12-14
torchx/Dockerfile: use newer pip for installing

d4l3k:
This makes the Dockerfile use the latest version of pip which has a better dependency resolver which fixes our build timeouts

Test plan:

CI

### DIFF
--- a/torchx/runtime/container/Dockerfile
+++ b/torchx/runtime/container/Dockerfile
@@ -2,6 +2,9 @@ FROM pytorch/pytorch:1.13.0-cuda11.6-cudnn8-runtime
 
 WORKDIR /app
 
+# upgrade pip to 22.x+ which has a faster dependency resolver
+RUN pip install --upgrade pip
+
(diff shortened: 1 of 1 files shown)


torchx/runtime/container/Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM pytorch/pytorch:1.13.0-cuda11.6-cudnn8-runtime
 
 WORKDIR /app
 
+# upgrade pip to 22.x+ which has a faster dependency resolver
```
kurman, line 5: Is there an existing image with upgraded pip? This could further reduce build time.

d4l3k: pip is pretty small -- upgrading is pretty fast

kurman (approved):

priyaramani (approved):

d4l3k:
GCP batch setup is broken right now, merging

PR #669, opened 2022-12-15
Add DDP component support in GCP Batch 

priyaramani:
Add DDP component support in GCP Batch:
- Update rank0_env to the newly added env var by GCP Batch: BATCH_MAIN_NODE_HOSTNAME
- Enable host level and docker level communication as they aren't configured by default in GCP container images (but that's the default behavior in other schedulers like AWS Batch)
- Update integration tests to run ddp style job and check logs
- Add ability in component integration tests to run gcp jobs but not running them yet as the service account needs to given permissions to push to container repos. Will enable them in a different PR.

Test plan:
Integration tests, CI

### DIFF
--- a/.github/workflows/gcp-batch-integration-tests.yaml
+++ b/.github/workflows/gcp-batch-integration-tests.yaml
@@ -30,14 +30,14 @@ jobs:
         run: |
           set -eux
           pip install -e .[dev]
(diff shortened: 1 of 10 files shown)


d4l3k (approved):
see comments in diffusion

PR #672, opened 2022-12-16
(torchx/aws_batch) Added username tag + make list API only return torchx submitted jobs by filtering on tag

kiukchung:
Two QoL improvements to aws_batch scheduler:

1. Added a username tag (torchx.pytorch.org/user) which is configurable via cfg.user (gets unix user by default)
2. Make `list` API only return torchx submitted jobs by filtering on the existence of `torchx.pytorch.org/version`. This make s it possible to user torchx on batch clusters that are shared with other jobs being launched from different launchers.

Future improvements:
1. Add filter to only list jobs for a particular user? --> requires the `cfg` to be plumbed through the list API for scheduler and runner.

### DIFF
--- a/torchx/schedulers/aws_batch_scheduler.py
+++ b/torchx/schedulers/aws_batch_scheduler.py
@@ -34,7 +34,7 @@ See
 https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html#cli-create-repository
 for how to create a image repository.
 """
-
+import getpass
 import re
 import threading
 from dataclasses import dataclass
@@ -78,6 +78,11 @@ from torchx.specs.api import (
 )
 from torchx.workspace.docker_workspace import DockerWorkspaceMixin
 from typing_extensions import TypedDict
+
+TAG_TORCHX_VER = "torchx.pytorch.org/version"
+TAG_TORCHX_APPNAME = "torchx.pytorch.org/app-name"
+TAG_TORCHX_USER = "torchx.pytorch.org/user"
+
 
 if TYPE_CHECKING:
     from docker import DockerClient
@@ -244,6 +249,7 @@ def _local_session() -> "boto3.session.Session":
 
 class AWSBatchOpts(TypedDict, total=False):
     queue: str
+    user: str
     image_repo: Optional[str]
     share_id: Optional[str]
     priority: Optional[int]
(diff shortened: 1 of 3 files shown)


torchx/schedulers/aws_batch_scheduler.py
```diff
@@ -455,6 +462,12 @@ def _cancel_existing(self, app_id: str) -> None:
     def _run_opts(self) -> runopts:
         opts = runopts()
         opts.add("queue", type_=str, help="queue to schedule job in", required=True)
+        opts.add(
+            "user",
+            type_=str,
+            default=getpass.getuser(),
+            help="The username to tag the job with. `getpass.getuser()` if not specified.",
+        )
```
d4l3k, line 470: I wonder if we want to add a more general concept of identity to TorchX that we could allow overriding with entrypoints etc but I think this is fine for now

torchx/schedulers/aws_batch_scheduler.py
```diff
@@ -582,36 +595,46 @@ def list(self) -> List[ListAppResponse]:
...
-                },
-            ],
+            filters=[EVERY_STATUS],
+            # describe-jobs API can take up to 100 jobIds
+            PaginationConfig={"MaxItems": 100},
         ):
-            job_summary_list.extend(resp["jobSummaryList"])
-        return job_summary_list
+
+            def get_torchx_submitted_jobs(job_ids: List[str]) -> List[ListAppResponse]:
```
d4l3k, line 617: can we define this outside of the forloop since it doesn't depend on any variables in the for loop?

torchx/schedulers/test/aws_batch_scheduler_test.py
```diff
@@ -127,7 +127,7 @@ def test_submit_dryrun_with_priority(self) -> None:
     def test_submit_dryrun(self) -> None:
         scheduler = create_scheduler("test")
         app = _test_app()
-        cfg = AWSBatchOpts({"queue": "testqueue"})
+        cfg = AWSBatchOpts({"queue": "testqueue", "user": "testuser"})
```
d4l3k, line 130: might be nice to have a separate test just for tags but not a big deal

d4l3k (approved):
LGTM -- some small comments

torchx/schedulers/test/aws_batch_scheduler_test.py
kiukchung: done

torchx/schedulers/aws_batch_scheduler.py
kiukchung: done

PR #673, opened 2022-12-20
(torchx/aws_batch) Enable Region Selection in TorchConfig

ashvinnihalani:
QOL improvements for support for AWS regions. 

Current Behavior: Use the default Region specified in .aws config

Proposed Behaviour: Have manual override in .torchxconfig. If no override and no default throw and error

Future Improvements: Add profile support. Add fargate support.

### DIFF
--- a/torchx/schedulers/aws_batch_scheduler.py
+++ b/torchx/schedulers/aws_batch_scheduler.py
@@ -55,6 +55,8 @@ from typing import (
 
 import torchx
 import yaml
+
+from botocore.exceptions import NoRegionError
 from torchx.schedulers.api import (
     AppDryRunInfo,
     DescribeAppResponse,
@@ -88,7 +90,7 @@ if TYPE_CHECKING:
     from docker import DockerClient
 
 JOB_STATE: Dict[str, AppState] = {
-    "SUBMITTED": AppState.PENDING,
+    "SUBMITTED": AppState.SUBMITTED,
     "PENDING": AppState.PENDING,
     "RUNNABLE": AppState.PENDING,
     "STARTING": AppState.PENDING,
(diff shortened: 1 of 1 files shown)


kiukchung:
can you make sure the lint, pyre, and python unittests pass?

torchx/schedulers/aws_batch_scheduler.py
```diff
@@ -588,13 +641,35 @@ def log_iter(
     def list(self) -> List[ListAppResponse]:
         # TODO: get queue name input instead of iterating over all queues?
         all_apps = []
-        for resp in self._client.get_paginator("describe_job_queues").paginate():
-            queue_names = [queue["jobQueueName"] for queue in resp["jobQueues"]]
-            for qn in queue_names:
-                all_apps.extend(self._list_by_queue(qn))
+        regions = _local_sessions_available_regions()
+        for region in regions:
```
kiukchung, line 645: have you measured how long this takes? last I tried this, it drastically worsens latency for m5 since we have a ton of job queues across regions. (by drastic I mean ~40seconds for the list API)

kiukchung (changes requested):
please add unittests for the changes and make sure all the CI checks pass

torchx/schedulers/aws_batch_scheduler.py
```diff
@@ -240,19 +244,36 @@ def wrapper() -> T:
...
+
+
 @_thread_local_cache
-def _local_session() -> "boto3.session.Session":
+def _local_session_region(region: Optional[str] = None) -> str:
     import boto3.session
 
-    return boto3.session.Session()
+    if region:
+        return boto3.session.Session(region_name=region).region_name
```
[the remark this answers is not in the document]
ashvinnihalani: The original point was to check if a region name had been cached or specified in the config file or environment variable which is handled by the 
`return boto3.session.Session().region_name` in line 260. 

Adding a region parameter for the method signature similar to _local_session. 

We can either remove the region parameter altogether or just return the region string without constructing a boto3.session object. 

Personally, I am in favor of just returning the region string, that way if the priority of the region specification changes in the future we can just change the implementation without breaking backward compat.

torchx/schedulers/aws_batch_scheduler.py
ashvinnihalani: Ill double-check for M5. However, did this even work across regions originally? A more apt comparison might be specifying the region in the .torchxconfig vs just using the other methods which I think I originally did.

kiukchung:
Thanks for the suggestion! Closing as stale - this has been unmerged since 2022 and the scheduler config surface has changed. A region override via .torchxconfig is still a reasonable ask; please re-spin against current main if you still need it.

PR #674, opened 2022-12-22
(torchx/schedulers) Remove redundantly setting dryruninfo._app and _cfg in k8s, gcp_batch, aws_batch, and docker scheduler subclasses since the scheduler abstract class sets it in the dryrun() API

kiukchung:
Test plan:


d4l3k (approved):
LGTM, nice catch!

PR #675, opened 2022-12-29
Update gcp batch lib version in setup.py

priyaramani:
As a preparation for release 0.4, update gcp batch lib version in setup.py to match the one used in dev-requirements 

Test plan:
CI


kurman (approved):
Approving.

PR #678, opened 2023-01-03
Find default namespace from kube_config.

liuwenchao:
For kubernetes scheduler, find default namespace from kube_config other than hard-coded "default" value.

Right now the default namespace for list method is hard-coded as "Default", which is not the case for multiple user scenario in kubernetes cluster. The default namespace can be read from current context of kube configuration.

Test plan:

Tested manually with a generated kube config.


d4l3k:
@liuwenchao thanks for the contribution! This looks like a great feature to have. It would be nice to read this value for the namespace to launch jobs from. 

Happy to accept this diff as is but looks like the lint is failing. Can you fix the lint and then I'll land?

Thanks!

liuwenchao:
@d4l3k Can you please point me how to reproduce the errors? scripts/lint.sh run successfully without error. 

```
$ scripts/lint.sh                                                                                        
Running linters ...
Checking torchx/tracker/backend/fsspec.py
Checking torchx/tracker/backend/test/fsspec_test.py
Checking torchx/workspace/docker_workspace.py
```

d4l3k:
Do you have the right version of the linters installed? Check the dev-requirements.txt file

Also the unit tests look like they're failing when there's no kubeconfig file present

liuwenchao:
@d4l3k please check if it's good now? Thank you~

d4l3k (approved):
LGTM -- it would be nice to add a unit test since we don't actually cover the Kubeconfig path in CI

liuwenchao:
@d4l3k It seems #679 has covered this part, so this PR is unnecessary now.

d4l3k:
@liuwenchao thanks for pointing that out, sorry I missed that before. I thought that was just for the mcad one but I guess not. Thanks for your help regardless :)

PR #679, opened 2023-01-05
TorchX-MCAD support

Sara-KS:
Add support to TorchX for the MCAD Kubernetes Scheduler

Test plan:
Created a Kubernetes MCAD scheduler test and added it to the TorchX test suite.

### DIFF
--- a/torchx/schedulers/kubernetes_mcad_scheduler.py
+++ b/torchx/schedulers/kubernetes_mcad_scheduler.py
@@ -36,6 +36,7 @@ Learn more about running distributed trainers :py:mod:`torchx.components.dist`
 import json
 import logging
 import re
+import subprocess
 
 import warnings
 from dataclasses import dataclass
@@ -54,9 +55,7 @@ from typing import (
 
 import torchx
 import yaml
-
-# ListAppResponse is a planned future update
-from torchx.schedulers.api import (  # noqa: F401 imported but unused
+from torchx.schedulers.api import (
     AppDryRunInfo,
     DescribeAppResponse,
     filter_regex,
@@ -82,7 +81,6 @@ from torchx.specs.api import (
     runopts,
     VolumeMount,
 )
-from torchx.specs.builders import make_app_handle
 
 # from torchx.workspace.docker_workspace import DockerWorkspace
 from torchx.workspace.docker_workspace import DockerWorkspaceMixin
@@ -366,7 +364,7 @@ def role_to_pod(
 
 def mcad_svc(svc_name: str, namespace: str, service_port: str) -> "V1Service":
     from kubernetes.client.models import (  # noqa: F401, F811
-        V1Container,  # noqa: F811 redefinition of unused
+        V1Container,
         V1ContainerPort,
         V1EmptyDirVolumeSource,
         V1EnvVar,
@@ -661,6 +659,15 @@ def get_role_information(genericItems: Iterable[Dict[str, Any]]) -> Dict[str, An
                             roles[role_name].num_replicas += 1
 
     return roles
+
+
+def get_appwrapper_status(app: Dict[str, str]) -> AppState:
+    if "status" in app.keys():
+        # pyre-fixme
+        return JOB_STATE[app["status"]["state"]]
+    else:
+        # Handle case where appwrapper is created but pending dispatch
+        return JOB_STATE["Pending"]
 
 
 # Does not handle not ready to dispatch case
@@ -1069,10 +1076,14 @@ class KubernetesMCADScheduler(DockerWorkspaceMixin, Scheduler[KubernetesMCADOpts
         else:
             return iterator
 
-    def list(self) -> List[str]:
-        job_names = []
-        # TO DO: retrieve correct namespace here
-        namespace = "default"
+    def list(self) -> List[ListAppResponse]:
+        p1 = subprocess.run(
+            ["kubectl", "config", "view", "--minify"],
+            stdout=subprocess.PIPE,
+            check=True,
+        )
+        namespace_id = p1.stdout.decode("utf-8").split().index("namespace:")
+        namespace = p1.stdout.decode("utf-8").split()[namespace_id + 1]
 
         resp = self._custom_objects_api().list_namespaced_custom_object(
             group="mcad.ibm.com",
@@ -1081,13 +1092,14 @@ class KubernetesMCADScheduler(DockerWorkspaceMixin, Scheduler[KubernetesMCADOpts
             plural="appwrappers",
             timeout_seconds=30,
         )
-        items = resp["items"]
-        for item in items:
-            name = item["metadata"]["name"]
-            app_id = f"{namespace}:{name}"
-            app_handle = make_app_handle("kubernetes_mcad", "default", app_id)
-            job_names.append(app_handle)
-        return job_names
+
+        return [
+            ListAppResponse(
+                app_id=f"{namespace}:{item['metadata']['name']}",
+                state=get_appwrapper_status(item),
+            )
+            for item in resp["items"]
+        ]
 
 
 def create_scheduler(session_name: str, **kwargs: Any) -> KubernetesMCADScheduler:
--- a/torchx/schedulers/kubernetes_mcad_scheduler.py
+++ b/torchx/schedulers/kubernetes_mcad_scheduler.py
@@ -1,0 +1,1030 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+
+This contains the TorchX Kubernetes scheduler which can be used to run TorchX
+components on a Kubernetes cluster.
+
+Prerequisites
+==============
+
+TorchX kubernetes scheduler depends on AppWrapper + MCAD and requires etcd intalled for distributed job execution.
+
+Install MCAD 
+
+.. Deploying Multi-Cluster-Application-Dispatcher
+
+    https://github.com/project-codeflare/multi-cluster-app-dispatcher/blob/main/doc/deploy/deployment.md
+
+TorchX uses `torch.distributed.run <https://pytorch.org/docs/stable/elastic/run.html>`_ to run distributed training.
+This requires the installation of etcd service on your kubernetes cluster:
+
+.. code:: bash
+
+    kubectl apply -f https://github.com/pytorch/torchx/blob/main/resources/etcd.yaml
+
+
+Learn more about running distributed trainers :py:mod:`torchx.components.dist`
+
+"""
+
+import json
+import logging
+import re
+import warnings
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Any, Dict, Iterable, List, Mapping, Optional, Tuple, TYPE_CHECKING, Union, cast
+
+import torchx
+import yaml
+from torchx.schedulers.api import (
+    AppDryRunInfo,
+    DescribeAppResponse,
+    filter_regex,
+    ListAppResponse,
+    Scheduler,
+    split_lines,
+    Stream,
+)
+from torchx.schedulers.ids import make_unique
+from torchx.specs.api import (
+    AppDef,
+    AppState,
+    BindMount,
+    CfgVal,
+    DeviceMount,
+    macros,
+    ReplicaState,
+    ReplicaStatus,
+    RetryPolicy,
+    Role,
+    RoleStatus,
+    runopts,
+    VolumeMount,
+    Resource,
+)
+from torchx.specs.builders import make_app_handle
+#from torchx.workspace.docker_workspace import DockerWorkspace
+from torchx.workspace.docker_workspace import DockerWorkspaceMixin
+from typing_extensions import TypedDict
+
+import sys
+
+if TYPE_CHECKING:
+    from docker import DockerClient
+    from kubernetes.client import ApiClient, CustomObjectsApi
+    from kubernetes.client.models import (  # noqa: F401 imported but unused
+        V1Container,
+        V1ContainerPort,
+        V1EnvVar,
+        V1Pod,
+        V1PodSpec,
+        V1ResourceRequirements,
+        V1Service,
+    )
+    from kubernetes.client.rest import ApiException
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+# Kubernetes reserves a small amount of resources per host for the system. For
+# TorchX we always assume the entire host is being requested so we adjust the
+# requested numbers account for the node reserved resources.
+#
+# https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/
+RESERVED_MILLICPU = 100
+RESERVED_MEMMB = 1024
+
+RETRY_POLICIES: Mapping[str, Iterable[Mapping[str, str]]] = {
+    RetryPolicy.REPLICA: [],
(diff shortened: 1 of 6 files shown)


torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+
+This contains the TorchX Kubernetes scheduler which can be used to run TorchX
+components on a Kubernetes cluster.
```
d4l3k, line 11: nit: update to refer to MCAD

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+Prerequisites
+==============
+
+TorchX kubernetes scheduler depends on AppWrapper + MCAD and requires etcd intalled for distributed job execution.
+
+Install MCAD 
+
+.. Deploying Multi-Cluster-Application-Dispatcher
+
+    https://github.com/project-codeflare/multi-cluster-app-dispatcher/blob/main/doc/deploy/deployment.md
```
d4l3k, line 22: just make this a link instead of a code block

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+
+
+def sanitize_for_serialization(obj: object) -> object:
+    from kubernetes import client
+
+    api = client.ApiClient()
+    return api.sanitize_for_serialization(obj)
+
+
+def role_to_pod(
```
d4l3k, line 173: Is it possible to share this method between the kubernetes scheduler and mcad? I see the env differences, anything else?

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+    pattern = r"[a-z0-9\-]"
+    return "".join(re.findall(pattern, data.lower()))
+
+
+def get_port_for_service(app: AppDef) -> str:
+    # Initialize port to default
+    port = "29500"
+
+    # look for rdzv_endpoint
+    rdzv = "--rdzv_endpoint"
```
d4l3k, line 431: this seems pretty hacky -- can we instead grab the ports from the port_map in the role?

https://github.com/pytorch/torchx/blob/main/torchx/components/dist.py#L203

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+            genericitem: Dict[str, Any] = {
+                "replicas": 1,
+                "generictemplate": pod,
+            }
+            # TODO: retry support here
+            if role.max_retries > 0:
+                genericitem["maxRetry"] = role.max_retries
+                genericitem["policies"] = RETRY_POLICIES[role.retry_policy]
+                msg = f"""
+Role {role.name} configured with restarts: {role.max_retries}.
```
d4l3k, line 500: nit: add comment about why we're warning about this "MCAD doesn't support retries. Role ... configured..."

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+    ]
+
+    my_env_var = [
+        V1EnvVar(
+            name=f"MCAD_{cleanup_str(role.name)}_{replica_id}_HOSTS".upper().replace(
+                "-", ""
+            ),
+            value=f"{unique_app_id}-{replica_id}.{unique_app_id}",
+        )
+        for replica_id in range(role.num_replicas)
```
d4l3k, line 317: nit: these environment variables aren't strictly needed, just one for the rank 0 host

may want to prefix it with TORCHX_MCAD so users don't think they're provided by the scheduler

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+    """
+
+    service_port = get_port_for_service(app)
+
+    if not (0 < int(service_port) <= 65535):
+        msg = """Warning: rdzv_endpoint set to invalid port number. Value must be between 1-65535, with torchx default = 29500. Try updating --rdzv_port argument to dist.ddp. Setting port to default = 29500"""
+        service_port = "29500"
+        warnings.warn(msg)
+
+    svc_obj = mcad_svc(unique_app_id, namespace, service_port)
```
d4l3k, line 519: what's this service used for? for pytorch rendezvous it's using the hostnames and not this right?

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+    mounts_key = "volumeMounts"
+
+    # resource keys
+    cpu_key = "cpu"
+    gpu_key = "gpu"
+    request_key = "requests"
+    mem_key = "memory"
+
+    for genericItem in genericItems:
+        if gt_key in genericItem.keys():
```
d4l3k, line 574: this is really nested, can we flatten a lot of these out with continues?

i.e.

```
if gt_key not in genericItem.keys():
   continue
gt_result = generic_item[gt_key]
...
```

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+
+
+# Helper functions for MCAD generic items information -> TorchX Role
+def get_role_information(genericItems: Iterable[Dict[str, Any]]) -> Dict[str, Any]:
+    # Store unique role information
+    roles = {}
+
+    # nested dictionary keys
+    # meta data information
+    gt_key = "generictemplate"
```
d4l3k, line 552: nit: all caps for constants GT_KEY

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+        "apiVersion": "mcad.ibm.com/v1beta1",
+        "kind": "AppWrapper",
+        "metadata": {"name": unique_app_id},
+        "spec": job_spec,
+    }
+    return resource
+
+
+# Helper functions for MCAD generic items information -> TorchX Role
+def get_role_information(genericItems: Iterable[Dict[str, Any]]) -> Dict[str, Any]:
```
d4l3k, line 546: nit: snake case

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+
+        namespace, name = app_id.split(":")
+
+        print("\033[100m" + "\n ks - log_iter \n" + "\033[0m")
+        print(f"{name}")
+
+        pod_name = cleanup_str(f"{name}-{k}")
+
+        print("\033[100m" + "\n ks - log_iter \n" + "\033[0m")
+        print(f"{pod_name}")
```
d4l3k, line 1056: nit: remove debug prints

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+            iterator = split_lines(resp)
+
+        if regex:
+            return filter_regex(regex, iterator)
+        else:
+            return iterator
+
+    def list(self) -> List[ListAppResponse]:
+        p1 = subprocess.run(
+            ["kubectl", "config", "view", "--minify"],
```
d4l3k, line 1081: this is a clever trick :)

Might be better to grab it from the kube config using the Python API however

https://github.com/kubernetes-client/python/issues/363#issuecomment-389722355

Would be nice to integrate this into the normal k8s scheduler as well

d4l3k (commented):
It's looking good! Excited to see this :)

Would be nice to share some of these helper methods/spec generation with the Volcano integration where possible but if it doesn't make sense that's fine

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+    ReplicaStatus,
+    Resource,
+    RetryPolicy,
+    Role,
+    RoleStatus,
+    runopts,
+    VolumeMount,
+)
+
+# from torchx.workspace.docker_workspace import DockerWorkspace
```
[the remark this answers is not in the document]
Sara-KS: Removed in commit: 65362828973a37624a6774a6da36a7460b4f6a89

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1121 @@
...
+
+TorchX uses `torch.distributed.run <https://pytorch.org/docs/stable/elastic/run.html>`_ to run distributed training.
+This requires the installation of etcd service on your kubernetes cluster:
+
+.. code:: bash
+
+    kubectl apply -f https://github.com/pytorch/torchx/blob/main/resources/etcd.yaml
+
+
+Learn more about running distributed trainers :py:mod:`torchx.components.dist`
```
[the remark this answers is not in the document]
Sara-KS: Removed in commit: 65362828973a37624a6774a6da36a7460b4f6a89

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Updated in commit: 65362828973a37624a6774a6da36a7460b4f6a89

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Updated in commit: 65362828973a37624a6774a6da36a7460b4f6a89

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Primary differences are the function arguments (Volcano options or MCAD options) and the addition of the environment variable. This also leads to differences in the pod specs and metadata (section on V1Pod).

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Restructured in commit: 2c4a0c91493960d11a25068ab7729ac5eaf25177

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Updated constants in commit: 2c4a0c91493960d11a25068ab7729ac5eaf25177

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: updated in commit: 2c4a0c91493960d11a25068ab7729ac5eaf25177

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: This service is used for distributed jobs to enable the PyTorch rendezvous. Since MCAD treats the jobs holistically, it needs to know the Kubernetes service information and pod information and treats them as the same "AppWrapped" job. For example, a distributed job that will run as two pods will be represented in a single "App wrapper" as a pod with 2 replicas and 1 service. This includes needing the port that will be used for the PyTorch rendezvous. The Kubernetes Volcano scheduler automatically creates and destroys a service for distributed PyTorch jobs as well. However, Volcano manages this differently and so TorchX does not need to be aware of the Kubernetes service.

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Distributed PyTorch jobs fail if the ports do not match, this is related to the conversation on the mcad_svc function below.

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Updated message in commit: bcb0ebe432db59f8ecca3c0ac77bd8bdc2d5a83a

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Updated prefix in commit: bcb0ebe432db59f8ecca3c0ac77bd8bdc2d5a83a

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Updated MCAD and Volcano schedulers to use Kube config for getting the namespace in both list functions. Commit: bcb0ebe432db59f8ecca3c0ac77bd8bdc2d5a83a

torchx/schedulers/kubernetes_mcad_scheduler.py
d4l3k: interesting -- can we still pull the port from the AppDef port map instead of introspecting like this? would be nice to support jobs that aren't using torch elastic

torchx/schedulers/kubernetes_mcad_scheduler.py
d4l3k: is the service set as the hostname on the workers? I'm just not quite understanding why a service is needed here if the pods can directly talk to each other.  The service is for load balancing/discovery but we're using the pod host names so if I understanding correctly shouldn't be needed

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -0,0 +1,1090 @@
...
+
+"""
+
+This contains the TorchX Kubernetes_MCAD scheduler which can be used to run TorchX
+components on a Kubernetes cluster via the Multi-Cluster-Application-Dispatcher (MCAD).
+
+Prerequisites
+==============
+
+TorchX kubernetes scheduler depends on AppWrapper + MCAD..
```
d4l3k, line 16: nit: extra period

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Reran tests and was able to update to use the role port_map. Commit: 697c5aa0b348cb38b6c79447aa70eda57f5fcbcb

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: This section just generates the yaml specs for a Kubernetes service. TorchX-Volcano generates a similar Kubernetes Service as well, but because Volcano handles this automation the rest of TorchX does not even need to be aware of the Kubernetes service. MCAD does not have this type of automation in place, so it had to be moved to the TorchX scheduler level and included in the generate yaml file that is evaluated by MCAD. If one were to simply comment out this mcad service and try to run a distributed job, the main worker stalls at waiting to rendezvous with all workers and the replica workers will reach a time out.

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: removed

d4l3k (approved):
LGTM, thanks for contributing!

Just waiting for CI to pass and then will merge

d4l3k:
docsbuild error seems unrelated, landing

Issue #680, opened 2023-01-11
Tracker with custom S3 (Minio)


ghpu:
## 🐛 Bug

Following the tracker example app, In fsspec_backend.conf, we can specify :
```
protocol=s3
root_path=s3://my-bucket
key=***
secret=**
```

In order to use Minio, one needs to pass client_kwargs with "endpointUrl" member as a struct : 
```
client_kwargs = {"endpointUrl":"http://myminio:9000"}
```

But the client_kwargs argument needs to be decoded as an object, not as a str : 
  File "/lib/python3.10/site-packages/s3fs/core.py", line 361, in set_session
    client_kwargs = self.client_kwargs.copy()
AttributeError: 'str' object has no attribute 'copy'

## Module (check all that applies):
 * [x ] `other` : tracker

## Environment info
```
PyTorch version: 1.13.1+cu117
Is debug build: False
CUDA used to build PyTorch: 11.7
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.10 (x86_64)
GCC version: (Ubuntu 12.2.0-3ubuntu1) 12.2.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.36

Python version: 3.10.7 (main, Nov 24 2022, 19:45:47) [GCC 12.2.0] (64-bit runtime)
Python platform: Linux-5.19.0-28-generic-x86_64-with-glibc2.36
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] botorch==0.6.0
[pip3] gpytorch==1.9.1
[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.24.1
[pip3] pytorch-lightning==1.5.10
[pip3] torch==1.13.1
[pip3] torch-model-archiver==0.7.0
[pip3] torchmetrics==0.10.3
[pip3] torchserve==0.7.0
[pip3] torchtext==0.14.1
[pip3] torchvision==0.14.1
[pip3] torchx==0.4.0
[conda] Could not collect

Versions of CLIs:
AWS CLI: N/A
gCloud CLI: None
AZ CLI: None
Slurm: N/A
Docker: 20.10.16, build 20.10.16-0ubuntu1
[...]

PR #681, opened 2023-01-12
(tracker/fsspec) Make fsspec tracker able to read nested kwargs in co…

kiukchung:
Resolves https://github.com/pytorch/torchx/issues/680

Test plan:

Added appropriate unittests

### DIFF
--- a/torchx/tracker/backend/fsspec.py
+++ b/torchx/tracker/backend/fsspec.py
@@ -11,7 +11,7 @@ import os
 import time
 from base64 import b32decode, b32encode
 from dataclasses import dataclass
-from typing import Iterable, Mapping, Optional
+from typing import Any, Dict, Iterable, Mapping, Optional
 
 import fsspec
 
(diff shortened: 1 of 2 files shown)


torchx/tracker/backend/fsspec.py
```diff
@@ -236,19 +236,33 @@ def __repr__(self) -> str:
         return f"<FsspecTracker: root_path={self._path_builder.path()}>"
 
 
-def _read_config(config_file: str) -> Mapping[str, str]:
+def _put_config(key: str, value: str, config: Dict[str, Any]) -> None:
```
kurman, line 239: This is a form of reduction, if you wish we can convert it into iterative approach something like this:

~~~
keys = key.split(".")

last_key = keys[-1]
parent_keys = keys[:-1]

config = {}
nested_config = config
for parent_key in parent_keys:
  if not parent_key: # takes care of any extra periods
    raise ValueError(f"Illegal config key `{key}`.")
  nested_config.setdefault(parent_key, {})
  nested_config = nested_config[parent_key]
nested_config[last_key] = value
~~~

kurman (commented):
Awesome thanks! Small note on non-recursive approach, let me know what you think.

kurman (approved):

torchx/tracker/backend/fsspec.py
kiukchung: is there a particular reason you prefer the iterative approach to the recursive one?

PR #686, opened 2023-01-20
torchx/slurm-integration-tests: use local slurm cluster instead of remote

d4l3k:
This creates a local slurm cluster using docker and docker-compose instead of using a remote cluster. This should reduce maintenance, make it easier for folks to test/contribute and make the cluster env fully reproducible.

This currently depends on a patch https://github.com/giovtorres/slurm-docker-cluster/pull/29

Test plan:

CI

spin up docker cluster

```
scripts/slurmdockerint.sh
```


onegigbyte (approved):
Nice.

cz-37 (approved):
lgtm!

Thank you very much!

PR #687, opened 2023-01-20
use minikube for integration tests

d4l3k:
This uses minikube for integration tests instead of running in a remote dedicated cluster. This makes the environment reproducible, reduces maintenance, and makes it easier for more users to contribute.

Test plan:

CI


kurman (approved):
Does this completely remove dependency on KFP?

d4l3k:
@kurman doesn't remove the kfp dependency yet, this is just for volcano rn

d4l3k:
CI green, landing

PR #688, opened 2023-02-09
Azure batch scheduler implementation

kurman:
1. Modify commands to initialize scheduler with options that can be defined in config. Generally most of the schedulers can operate using scheduler options, however in some cases for multi-tenant setup operations needs to be pre-configured. For example, Azure Batch allows multiple Batch accounts
2. Initial implementation of the Azure batch scheduler. Provides support for scheduling, stopping, listing and describing jobs. Azure Batch has a good support HPC support and training jobs should work out of the box.

Major things that needs to be addressed:
- Support for patching using Docker (currently it can run existing image)
- Creating autoscaling pools that maps to resource requirements

Testing
=======
- Unit test
- `python -m torchx.cli.main run --scheduler azure_batch --scheduler_args image_repo=ghcr.io/pytorch/torchx utils.echo --image alpine:latest  --msg hello`

Screenshot 2023-02-08 at 7 15 27 PM
Screenshot 2023-02-08 at 7 15 43 PM
Screenshot 2023-02-08 at 7 15 51 PM

### DIFF
--- a/torchx/schedulers/azure_batch_scheduler.py
+++ b/torchx/schedulers/azure_batch_scheduler.py
@@ -1,0 +1,338 @@
+#!/usr/bin/env python3
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+
+Initial support for Azure Batch scheduler.
+
+This scheduler is in prototype stage and may change without notice.
+
+Overview
+========
+https://learn.microsoft.com/en-us/azure/batch/
+
+
+Prerequisites
+==============
+
+You'll need to setup:
+- Azure Batch Account+credentials.
+- Batch Pool, that:
+    - has enough resources, or
+    - supports autoscaling (https://learn.microsoft.com/en-us/azure/batch/batch-automatic-scaling)
+    - VMs that support container workloads (https://learn.microsoft.com/en-us/azure/batch/batch-docker-container-workloads)
+
+Configuration
+=============
+Use .torchxconfig to configure the Batch Account information
+
+.. code-block:: ini
+[scheduler:azure_batch]
+batch_account_name = <ACCOUNT NAME>
+batch_account_key = "<ACCOUNT_KEY>"
+batch_account_url: <ACCOUNT_URL>
+.. code-block:: ini
+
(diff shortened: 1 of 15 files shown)


torchx/schedulers/azure_batch_scheduler.py
```diff
@@ -0,0 +1,338 @@
...
+            mounts: false
+            elasticity: false
+    """
+
+    def __init__(
+        self,
+        session_name: str,
+        batch_account_name: Optional[str],
+        batch_account_key: Optional[str],
+        batch_account_url: Optional[str],
```
kiukchung, line 146: why do these have to be taken in the constructor arguments (versus via the runoptions)? Looks like they are used when building the client in `_build_client()` and the client is create on-demand in the `schedule` API (rather than it being created once and being cached somewhere).

It looks like the real underlying reason why you added these to the constructor arguments is b/c the runcfg is not passed to the other APIs (e.g. list, describe, etc). Why not make the runcfg available to the other APIs instead of splitting scheduler configs between constructor args and runoptions?

torchx/schedulers/azure_batch_scheduler.py
```diff
@@ -0,0 +1,338 @@
...
+    ) -> None:
+        super().__init__("azure_batch", session_name)
+        self.batch_account_name = batch_account_name
+        self.batch_account_key = batch_account_key
+        self.batch_account_url = batch_account_url
+
+    def _build_client(self) -> BatchServiceClient:
+        credentials = SharedKeyCredentials(
```
kiukchung, line 154: fwiw, for AWS, the boto3 client grabs the default credentials (as configured in `~/.aws/credentials` or from instance profile (if on an EC2 instance). Does a similar concept not exist in Azure? If it does we should follow the same strategy to deal with credentials.

For example, if a user using `aws_batch` scheduler wanted to change to a different credential (either to access a different account or to switch to a different IAM role), they would change the creds provider in the `~/.aws` configuration rather than having to pass this information to torchx. This is much clearer to the users since they can use the vanilla AWS credentials chain  rather than having to pass parameters to torchx.

kiukchung:
@kurman any plans to actually commit this?

kiukchung:
Closing this long-stale branch - scheduler extensions are now served by the torchx.plugins architecture (schedulers register as plugins rather than in-tree), so an Azure Batch integration would be built as a plugin against current main.

PR #693, opened 2023-02-24
Torchx mcad coscheduler

Sara-KS:
This addition enables the option of running TorchX-MCAD with a Kubernetes co-scheduler and PodGroups. In shared Kubernetes clusters where some users do not represent their jobs with AppWrappers, the additional use of a co-scheduler helps ensure correct gang scheduling. 

Test plan:
Updated CI test coverage
Testing includes using the new`--coscheduler_name` flag for the kubernetes_mcad scheduler and ensuring that Kubernetes or OpenShift reports the named scheduler when pods are scheduled.

### DIFF
--- a/torchx/schedulers/kubernetes_mcad_scheduler.py
+++ b/torchx/schedulers/kubernetes_mcad_scheduler.py
@@ -512,7 +512,6 @@ def app_to_resource(
                 image_secret,
                 coscheduler_name,
             )
-            # pod.metadata.labels.update(pod_labels(app, role_idx, role, replica_id))
             pod.metadata.labels.update(
                 pod_labels(
                     app, role_idx, role, replica_id, coscheduler_name, unique_app_id
--- a/torchx/schedulers/kubernetes_mcad_scheduler.py
+++ b/torchx/schedulers/kubernetes_mcad_scheduler.py
@@ -168,6 +168,7 @@ def role_to_pod(
     role: Role,
     service_account: Optional[str],
     image_secret: Optional[str],
+    coscheduler_name: Optional[str],
 ) -> "V1Pod":
     from kubernetes.client.models import (  # noqa: F811 redefinition of unused
         V1Container,
@@ -338,6 +339,7 @@ def role_to_pod(
             service_account_name=service_account,
             volumes=volumes,
             node_selector=node_selector,
+            scheduler_name=coscheduler_name,
         ),
         metadata=V1ObjectMeta(
             name=name,
@@ -466,8 +468,8 @@ def app_to_resource(
     namespace: str,
     service_account: Optional[str],
     image_secret: Optional[str],
+    coscheduler_name: Optional[str],
     priority: Optional[int] = None,
-    coscheduler: Optional[bool] = False,
 ) -> Dict[str, Any]:
     """
     app_to_resource creates a AppWrapper/MCAD Kubernetes resource definition from
@@ -480,7 +482,7 @@ def app_to_resource(
 
     unique_app_id = cleanup_str(make_unique(app.name))
 
-    if coscheduler:
+    if coscheduler_name is not None:
         genericitem_pod_group = create_pod_group(app, namespace, unique_app_id)
         genericitems.append(genericitem_pod_group)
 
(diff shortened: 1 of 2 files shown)


torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -352,6 +353,31 @@ def role_to_pod(
     )
 
 
+def create_pod_group(role: Role, namespace: str, app_id: str) -> "Dict[str, Any]":
+    pod_group_name = app_id + "-" + cleanup_str(role.name) + "-pg"
+
+    pod_group: Dict[str, Any] = {
```
d4l3k, line 359: Would be nice to add a  link to the schema definition for this type. I think it's https://github.com/kubernetes-sigs/scheduler-plugins/blob/56e2398e5051f796d5f02adb821d57eb8b6c20a7/config/crd/bases/scheduling.sigs.k8s.io_podgroups.yaml#L9 ?

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -473,8 +506,13 @@ def app_to_resource(
                 replica_role,
                 service_account,
                 image_secret,
+                coscheduler_name,
+            )
+            pod.metadata.labels.update(
+                pod_labels(
+                    app, role_idx, role, replica_id, coscheduler_name, unique_app_id
```
d4l3k, line 513: there's a lot of fields here. Think it's better to switch these to use `app=app, coscheduler_name=coscheduler_name, ...` to avoid accidentally passing the wrong field to the wrong param

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -861,9 +900,15 @@ def _submit_dryrun(
...
         assert isinstance(namespace, str), "namespace must be a str"
 
+        coscheduler_name = cfg.get("coscheduler_name")
+        assert coscheduler_name is None or isinstance(
+            coscheduler_name, str
+        ), "coscheduler_name must be a string"
+
         resource = app_to_resource(
-            app, namespace, service_account, image_secret, priority
+            app, namespace, service_account, image_secret, coscheduler_name, priority
```
d4l3k, line 909: likewise kwargs for this

d4l3k (approved):

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Included the PodGroup CRD (release 1.24) in the updated documentation.

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Updated

kiukchung:
> 2/27/2023 Component Integration Tests/ Kubernetes Dist Train Integration Tests appear to be failing due to a missing environment variable: torchx/scritps/integ_test_utils.py expects `CONTAINER_REPO`. Current test failures indicate that the container repository information is not detected in the test environment and attempts a docker push with ":tag" format instead of "repo:tag". Related issue: #694

Hi @Sara-KS thanks for taking a look I actually fixed that particular issue yesterday (see: https://github.com/pytorch/torchx/blob/main/scripts/kube_dist_trainer.py#L45) but the `Kubernetest Dist Train Integration Test` is failing for a different reason (401 Unauthorized when querying for the status of the job): https://github.com/pytorch/torchx/actions/runs/4277421091/jobs/7446219941

Since I've left Meta and don't have access to TorchX's AWS account I'm unable to login and take a deeper look. Perhaps @kurman or @priyaramani  can help here.

Long term, @d4l3k and I have discussed moving that integ test to run directly on the CI runner with minikube instead of hitting a deployed EKS cluster. Tristan's already made `components integ test` and `kfp integ test` to do this, we just haven't gotten around making the same change for the k8s integ test.

FWIW `Components Integration Tests` should be passing now (if you rebase on top of main)

d4l3k:
That integration test doesn't use the mcad scheduler anyways so you can just ignore it for this PR

d4l3k:
LGTM Failures look like an oversight on my part on how tagging is handled for forks

Issue #261, opened 2021-10-15
[file-linter] Refactor file-linter and components-finder


aivanou:
File linter and components finder modules reside under `torchx.specs` module. Components finder module has runtime dependency on the `torchx.components` module: during `find` execution, components finder would dynamically load the module and all its submodule to search components. This creates circular dependency between `torchx.specs` and `torchx.components`, since `torchx.componetns` depends on the `torchx.specs`.

We need to move these modules out of the `torchx.specs` , and extract hardcoded module search parameter from the components finder

Issue #694, opened 2023-02-24
Kubernetes/KFP integration tests failing


d4l3k:
## 🐛 Bug

Module (check all that applies):
 * [ ] `torchx.spec`
 * [ ] `torchx.component`
 * [ ] `torchx.apps`
 * [ ] `torchx.runtime`
 * [ ] `torchx.cli`
 * [ ] `torchx.schedulers`
 * [ ] `torchx.pipelines`
 * [ ] `torchx.aws`
 * [ ] `torchx.examples`
 * [x] `other`

## To Reproduce

https://github.com/pytorch/torchx/commit/99ab9170f1e673c9ab3aa1e2776879a0d9929ee5

It's interesting that the Kubernetes dist train ones and KFP are failing but the Kubernetes component integration tests are passing.

Remediating this might be best to finish #685 and https://github.com/pytorch/torchx/pull/687

cc @kurman , @priyaramani who can access the KFP cluster

kiukchung:
Closing as not planned: the Kubeflow Pipelines integration and its workflow have been removed from the repository, so this failure no longer has anything to reproduce against.

PR #695, opened 2023-02-26
(torchx/components) Fix entrypoint loading to deal with deferred load…

kiukchung:
Fixes a bug + enhances custom component registration.

1. [bugfix] torchx not allowing components to be registered via `[torchx.components]` entrypoint because of a change in `utils.entrypoints.load_entrypoints()` that defers the entrypoint load by wrapping the entrypoints with a `_defer_load_fn()` but assumes that all entrypoints are functions (e.g. `foo.bar:baz`) and does not handle modules (e.g. `foo.bar`). Adds graceful handling of module entrypoints.

2. [enhancement] Allows users to specify an "empty" group (e.g. no prefix) for component names by using the `_*` "special" prefix in the alias key (similar to "hidden" variables in python). Example
    ```
    [torchx.components]
     _0 = torchx.components.dist
     _1 = torchx.components.utils
    ``` 
    names `dist.ddp` as just `ddp` and `utils.python` as just `python`
3. [change] If custom components are registered via entrypoints, skips loading builtins. Users can still load the builtins by adding an extra line in their entrypoint as:
    ```
    [torchx.components]
    foo = bar.baz
    _ = torchx.components
    ```
   > NOTE: this is a BC incompatible change, HOWEVER given that component registration via entrypoints was broken, I'm assuming that no one was registering their own components.

4. [docs] Reflected the changes above in the docs page

5. [todo fix] resolves https://github.com/pytorch/torchx/issues/261 since we no longer have `components_base.py `file
Test plan:
----------

Added unittests + revised unittests for `utils.entrypoints` and `specs.finder`

### DIFF
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -156,7 +156,7 @@ resource can then be used in the following manner:
 
 Registering Custom Components
 -------------------------------
-It is possible to author and register a custom set of components with the
+You can author and register a custom set of components with the
 ``torchx`` CLI as builtins to the CLI. This makes it possible to customize
 a set of components most relevant to your team or organization and support
 it as a CLI ``builtin``. This way users will see your custom components
@@ -166,8 +166,25 @@ when they run
 
  $ torchx builtins
 
-Custom components can be registered via the following modification of the ``entry_points``:
+Custom components can be registered via ``[torchx.components]`` entrypoints.
+If ``my_project.bar`` had the following directory structure:
 
+::
+
+ $PROJECT_ROOT/my_project/bar/
+     |- baz.py
+
+And ``baz.py`` had a single component (function) called ``trainer``:
+
+::
+
+ # baz.py
+ import torchx.specs as specs
+
+ def trainer(...) -> specs.AppDef: ...
+
+
+And the entrypoints were added as:
 
 .. testcode::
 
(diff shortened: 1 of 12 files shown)


d4l3k:
@kiukchung lots of CI failures (docs, pyre, unit) are the relevant ones I think

docs/source/advanced.rst
```diff
@@ -179,29 +196,95 @@ Custom components can be registered via the following modification of the ``entr
...
-::
+.. code-block:: shell-session
 
- $PROJECT_ROOT/my_project/bar/
-     |- baz.py
+ $ torchx run foo.baz.trainer -- --name "test app"
+
+
+When you register your own components, TorchX will not include its own builtins. To add TorchX's
+builtin components you must specify another entry as:
```
d4l3k, line 225: may want to mention you can do this in the config as well?

docs/source/advanced.rst
```diff
@@ -179,29 +196,95 @@ Custom components can be registered via the following modification of the ``entr
...
+
+
+.. testcode::
+
+   # setup.py
+   ...
+   entry_points={
+       "torchx.components": [
+           "foo = my_project.bar",
+           "torchx = torchx.components",
```
d4l3k, line 235: this now works recursively? got it, interesting OK

docs/source/advanced.rst
```diff
@@ -179,29 +196,95 @@ Custom components can be registered via the following modification of the ``entr
...
+For example:
+
+.. testcode::
+
+   # setup.py
+   ...
+   entry_points={
+       "torchx.components": [
+           "_0 = my_project.bar",
+           "_1 = torchx.components",
```
d4l3k, line 275: do we want to drop the dist. and utils. prefixes for TorchX components now? not a blocker for this PR

d4l3k (approved):
LGTM

docs/source/advanced.rst
kiukchung: it was always loading recursively (by file not by module though)

docs/source/advanced.rst
kiukchung: the one supported by `.torchxconfig` is super weird (I want to deprecate it in favor of supporting "alias" -> "module_name" style custom components from `.torchxconfig`). Currently what you can do is to have a section like

```
# .torchxconfig
[torchx.file]
get_file_contents = my.custom.file_reader:read
```

And when you run

```
torchx run my/custom.py:component_fn
```

We call `my.custom.file_reader.read("my/custom.py")` which we expect to load the `component_fn` into the global namespace. Took me a while to wrap my head around it, so I don't expect anyone to actually use it.

Instead I'd like to support something like this from `.torchxconfig`:

```
# .torchxconfig
[torchx.components]
foo = my.custom.component
_ = torchx.components
```

which would be the same as doing

```
# setup.py

entrypoints = {
  "torchx.components": [
        "foo = my.custom.component",
        "_ = torchx.components",
   ],
}
```

docs/source/advanced.rst
kiukchung: yeah so we could drop the "module" name from the builtin components (as discussed offline we'd have to keep the alias `dist.*`,`util.*` etc while hiding them from `torchx builtins`). Should be easy to do now since we can make the builtins load:

```
ModuleComponentsFinder("torchx.components.dist", group="")
ModuleComponentsFinder("torchx.components.utils", group="")
# ... and so on...
```

kiukchung:
Also resolves: https://github.com/pytorch/torchx/issues/694

PR #696, opened 2023-02-26
kfpint: switch to using minikube cluster

d4l3k:
This runs the KFP integration tests using a local minikube cluster instead of the old AWS cluster.

This disables the interpret(captum) and torchserve integration tests since those are primarily examples and I don't think anyone is using the torchserve integration. We do have unit tests for torchserve component.

This uses a custom fsspec filesystem `torchx_minio` as s3fs requires custom params.

Test plan:

```
scripts/setup_minikube.sh
scripts/setup_kfp.sh
INTEGRATION_TEST_STORAGE=torchx_minio://torchx/tests KFP_NAMESPACE=kubeflow scripts/kfpint.py --container_repo localhost:5000/torchx
```

### DIFF
--- a/.github/workflows/kfp-integration-tests.yaml
+++ b/.github/workflows/kfp-integration-tests.yaml
@@ -8,34 +8,8 @@ on:
 
 jobs:
   kfp-launch:
-    runs-on: ubuntu-20.04
-    permissions:
-      id-token: write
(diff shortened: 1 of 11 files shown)


kiukchung:
LGTM one less dep to AWS resources in CI!

PR #698, opened 2023-03-01
(torchx/tracker) Allow AppRun.run_from_env() to be used without launc…

kiukchung:
Currently, due to the fact that `AppRun.run_from_env()` (aliased as `torchx.tracker.app_run_from_env()`) hard-checks
`TORCHX_JOB_ID` env var to set the `job_id` field of the returned `AppRun` object, the application that uses the tracker HAS to be launched with torchx or the user has to manually set the `TORCHX_JOB_ID` to a dummy value. This makes the app tightly coupled with torchx launcher. 

This change does away with this limitation and allows apps to be written to use the tracker without having to be launched by torchx. When not launched by torchx `TORCHX_JOB_ID` won't be set and `run_from_env()` will create an empty AppRun with no trackers and the `job_id` field set to the name of the program.

Test plan:

Added unittest.

### DIFF
--- a/torchx/tracker/api.py
+++ b/torchx/tracker/api.py
@@ -240,13 +240,39 @@ class AppRun:
     @staticmethod
     @lru_cache(maxsize=1)  # noqa: B019
     def run_from_env() -> AppRun:
-        """Factory method do build AppRun that uses environment variabes to build it.
+        """
+        Creates an :py:class:`AppRun` from environment variables. The environment variables are set by
+        the torchx runner. Hence if the application is launched via the torchx CLI as:
 
-        Single instance of AppRun will be available for the duration of the application, thus expect that calling
-        this method will retorn same AppRun and hence same instances of the backend implementations.
+        .. code-block:: shell-session
+
+            $ torchx run utils_python --script main.py
+
+
+        And the tracker settings are configured in ``.torchxconfig``, then this function returns an
+        ``AppRun`` with the configured tracker backends. This function returns a singleton ``AppRun``
+        hence the same instances of the tracker backend objects.
+
+        .. note::
+                When the application is NOT launched via torchx, this function
+                will return an "empty" ``AppRun`` with the ``job_id`` set to a constant
+                ``<UNDEFINED>`` since the app was not "launched" (e.g. submitted as a job)
+                and hence no canonical ``job_id`` exists.
+                No trackers are hooked up to the ``AppRun`` hence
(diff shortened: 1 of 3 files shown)


torchx/tracker/api.py
```diff
@@ -246,7 +248,7 @@ def run_from_env() -> AppRun:
         this method will retorn same AppRun and hence same instances of the backend implementations.
         """
 
-        torchx_job_id = os.environ["TORCHX_JOB_ID"]
+        torchx_job_id = os.getenv("TORCHX_JOB_ID", default=Path(sys.argv[0]).name)
```
d4l3k, line 251: In any multi-user environment this seems problematic since most jobs will have the job ID `python` right? Is it better to randomly generate an ID here instead? or require users to manually specify one if not present? Seems like a sharp edge

I think adding a new `run_id: Optional[str]` param to run_from_env() makes more sense than just defaulting it

torchx/tracker/api.py
```diff
@@ -246,7 +248,7 @@ def run_from_env() -> AppRun:
         this method will retorn same AppRun and hence same instances of the backend implementations.
```
d4l3k, line 248: would be good to document the ID behavior

d4l3k (changes requested):

torchx/tracker/api.py
kiukchung: the issue with adding the `run_id` param to `run_from_env()` is that then it makes the app a bit confusing to read. Presumably I have something like this now in my code:

```
def my_main() -> None:
    app_run tracker.app_run_from_env("my-custom-run")
```

If I run the above as `python my_main.py` then the job-id is `"my-custom-run"` but if I launch it via torchx it becomes something like `local_cwd://torchx/utils-python_my_main-19ksjkc`

torchx/tracker/api.py
kiukchung: done

torchx/tracker/api.py
d4l3k: I'd say the manual specified one should always take precedence

You could also construct it directly and not use env variables. But still defaulting the job id to "python" seems wrong

kiukchung: per our offline discussion, updated the job_id to `"<UNDEFINED>"` when the job is not launched from torchx since there is no canonical job_id.

d4l3k (approved):
LGTM

PR #699, opened 2023-03-06
Kubernetes Recommended Labels

Sara-KS:
Kubernetes recommends a set of standard labels for Kubernetes Objects. By updating some of the current TorchX Pod, Service, and PodGroup labels to match the recommended Kubernetes labels, this PR aims to improve the interoperability of TorchX on Kubernetes with common cloud tools. For more information on recommended Kubernetes labels, see https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

Test plan:
Updated the CI tests for kubernetes_mcad, kubernetes, and impacted KubeFlow Pipelines adapter

### DIFF
--- a/torchx/pipelines/kfp/test/adapter_test.py
+++ b/torchx/pipelines/kfp/test/adapter_test.py
@@ -129,11 +129,13 @@ outputs: []
             self.assertEqual(
                 a.pod_labels,
                 {
-                    "torchx.pytorch.org/version": torchx.__version__,
-                    "torchx.pytorch.org/app-name": "test",
+                    "app.kubernetes.io/instance": "test",
+                    "app.kubernetes.io/name": "test",
+                    "app.kubernetes.io/part-of": "torchx.pytorch.org",
+                    "app.kubernetes.io/version": torchx.__version__,
+                    "torchx.pytorch.org/replica-id": "0",
                     "torchx.pytorch.org/role-index": "0",
                     "torchx.pytorch.org/role-name": "trainer",
-                    "torchx.pytorch.org/replica-id": "0",
                 },
             )
 
--- a/torchx/pipelines/kfp/test/adapter_test.py
+++ b/torchx/pipelines/kfp/test/adapter_test.py
@@ -130,12 +130,13 @@ outputs: []
                 a.pod_labels,
                 {
                     "app.kubernetes.io/instance": "test",
+                    "app.kubernetes.io/managed-by": "torchx.pytorch.org",
                     "app.kubernetes.io/name": "test",
-                    "app.kubernetes.io/part-of": "torchx.pytorch.org",
-                    "app.kubernetes.io/version": torchx.__version__,
-                    "torchx.pytorch.org/replica-id": "0",
+                    "torchx.pytorch.org/version": torchx.__version__,
+                    "torchx.pytorch.org/app-name": "test",
                     "torchx.pytorch.org/role-index": "0",
                     "torchx.pytorch.org/role-name": "trainer",
+                    "torchx.pytorch.org/replica-id": "0",
                 },
             )
 
--- a/torchx/pipelines/kfp/adapter.py
+++ b/torchx/pipelines/kfp/adapter.py
@@ -175,7 +175,7 @@ def component_from_app(
                 ),
             )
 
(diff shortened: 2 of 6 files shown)


torchx/pipelines/kfp/test/adapter_test.py
```diff
@@ -129,11 +129,13 @@ def pipeline() -> None:
             self.assertEqual(
                 a.pod_labels,
                 {
-                    "torchx.pytorch.org/version": torchx.__version__,
-                    "torchx.pytorch.org/app-name": "test",
+                    "app.kubernetes.io/instance": "test",
+                    "app.kubernetes.io/name": "test",
+                    "app.kubernetes.io/part-of": "torchx.pytorch.org",
+                    "app.kubernetes.io/version": torchx.__version__,
```
d4l3k, line 135: Does this make sense to set as the torchx version? While TorchX is used to launch the job we usually use "app" to refer to the user running code so to stay in line with TorchX terminology this should be the user app/model version instead

What kind of tooling are you using with this? I'm wondering how this information gets surfaced to users etc in practice

d4l3k:
The monitor unit test is known to be flaky -- I'm rerunning the mac build but that's not anything to worry about

torchx/pipelines/kfp/test/adapter_test.py
Sara-KS: Thank you for the feedback. Previously, this label from the kubernetes scheduler was key `LABEL_VERSION = "torchx.pytorch.org/version"` and value `torchx.__version__` so I kept the same information but split it into separate pieces of information `app.kubernetes.io/part-of` and `app.kubernetes.io/version`. 

To match the TorchX terminology, would it make more sense to use model version here? If so, can you point me to how that information is stored? Alternatively, I could add an additional label that is more application specific such as `torchx.pytorch.org/model-version = model version`. This would be more in line with other torchx specific labels such as the `torchx.pytorch.org/role-index`. 

Regarding tooling, Kubernetes labels can be surfaced to users in a variety of ways. Some examples include:
- kubectl and Openshift CLI support showing object labels
- Red Hat OpenShift Console - all object labels are displayed (unless configured otherwise)
- Datadog - [extracts labels as tags](https://docs.datadoghq.com/containers/kubernetes/tag/?tab=containerizedagent)
- CodeFlare - we are planning to support TorchX-MCAD as an option through the[ CLI] (https://github.com/project-codeflare/codeflare-cli), which will make use of Kubernetes object labels


Additional reference for Kubernetes Labels:
Best practices -[section 3 ](https://www.redhat.com/sysadmin/kubernetes-labels-best-practices )

d4l3k: We don't have a strong concept of model version right now -- the closest thing would be the base docker tag or maybe info from setup.py or git.

I'm not opposed to making this change just worry it could be a tad confusing since there's more than one thing that matters here. 

How specifically do you anticipate the app.kubernetes.io/part-of tags being used? Wondering if there's specific tooling/need that requires that format instead of  the current torchx specific label. I guess you can filter by the label in UIs etc to find all TorchX jobs?

TorchX releases are relatively infrequent so I don't expect users need to check the version that often

Do you think `app.kubernetes.io/managed-by: torchx` makes more sense than part-of? Or does MCAD set the managed-by tag?

Sara-KS: Hello,

I had some discussions with our tooling team around this today. They mainly need some of the kubernetes standard labels for queuing, filtering, and correlation. Instead of touching the existing pod labels, would it be preferable to keep the current torchx specific labels and add the following 3 standard labels:
```
"app.kubernetes.io/name": application name
"app.kubernetes.io/instance" : unique application name
"app.kubernetes.io/managed-by": torchx
```

d4l3k: @Sara-KS that seems quite reasonable to me

Sara-KS: @d4l3k I reverted back to the original pod labels and then added the three standard labels described above. Commits dc5bfa8618023e9c7f8918f9d776ff4c41487298 and 2fd40a946c4ce0f55b6776adb0cdd79471b47e37

d4l3k (approved):
LGTM, thanks for updating!

Issue #700, opened 2023-03-08
[torchx/schedulers][aws_batch] Unable to submit jobs larger than 5 nodes


kiukchung:
## 🐛 Bug

Cannot run more than 5 node job using torchx + aws_batch.

This happens because torchx [creates a `NodeProperty` for each node replica](https://github.com/pytorch/torchx/blob/main/torchx/schedulers/aws_batch_scheduler.py#L417) (so 6 `NodeProperties` if `-j 6x[nproc_per_node]`). But aws batch seems to have an [UNDOCUMENTED](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html#nodeProperties) limit on the number of node properties in the `RegisterJobDefinition` API.

Module (check all that applies):
 * [ ] `torchx.spec`
 * [ ] `torchx.component`
 * [ ] `torchx.apps`
 * [ ] `torchx.runtime`
 * [ ] `torchx.cli`
 * [x] `torchx.schedulers`
 * [ ] `torchx.pipelines`
 * [ ] `torchx.aws`
 * [ ] `torchx.examples`
 * [ ] `other`

## To Reproduce

```
torchx run -s aws_batch dist.ddp -j 6x1 --script my_script.py
```

Errors out with 
```
Traceback (most recent call last):
  File "/home/ubuntu/.pyenv/versions/ape/bin/torchx", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/.pyenv/versions/3.9.15/envs/ape/lib/python3.9/site-packages/torchx/cli/main.py", line 115, in main
    run_main(get_sub_cmds(), argv)
  File "/home/ubuntu/.pyenv/versions/3.9.15/envs/ape/lib/python3.9/site-packages/torchx/cli/main.py", line 111, in run_main
    args.func(args)
  File "/home/ubuntu/.pyenv/versions/3.9.15/envs/ape/lib/python3.9/site-packages/torchx/cli/cmd_run.py", line 248, in run
    self._run(runner, args)
  File "/home/ubuntu/.pyenv/versions/3.9.15/envs/ape/lib/python3.9/site-packages/torchx/cli/cmd_run.py", line 209, in _run
    app_handle = runner.run_component(
  File "/home/ubuntu/.pyenv/versions/3.9.15/envs/ape/lib/python3.9/site-packages/torchx/runner/api.py", line 190, in run_component
    return self.schedule(dryrun_info)
[...]

PR #702, opened 2023-03-11
(torchx/schedulers)(bugfix) Make it possible to launch jobs onto aws_…

kiukchung:
Fixes: https://github.com/pytorch/torchx/issues/700

See issue for details on why the aws_batch scheduler failed to run jobs with more than 5 nodes.

> NOTE: this change makes it in such a way that the `macro.replica_id`  DOES NOT resolve in ENV vars. 
     this isn't quite possible unless we wrap the command with an env resolver function in bash.
     Given that it was between disabling the replica_id macro for env vars versus not being able to submit jobs larger than 5 nodes, I figured the later was more important to unblock right now, and forward fix the macro resolution for env vars.

Test plan:

Unittests

### DIFF
--- a/torchx/schedulers/aws_batch_scheduler.py
+++ b/torchx/schedulers/aws_batch_scheduler.py
@@ -37,8 +37,9 @@ for how to create a image repository.
 import getpass
 import re
 import threading
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 from datetime import datetime
+from enum import auto, Enum
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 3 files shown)


torchx/schedulers/aws_batch_scheduler.py
```diff
@@ -252,7 +338,7 @@ class AWSBatchOpts(TypedDict, total=False):
     user: str
     image_repo: Optional[str]
     share_id: Optional[str]
-    priority: Optional[int]
+    priority: int
```
d4l3k, line 341: why is priority not-optional?

d4l3k (commented):
lgtm

torchx/schedulers/aws_batch_scheduler.py
kiukchung: it defaults to 0 in _runopts(). Priority isn't actually used if share_id is None

PR #703, opened 2023-03-14
(torchx/schedulers) AWSBatchScheduler; add privileged option to enabl…

kiukchung:
1. Adds privileged runopt to aws batch scheduler
2. Also fixes a double call to create trackers from env in `tracker/api.py`, which discards parent run source information since a new set of trackers are created and hooked to the AppRun.

Test plan:

Added unittests to cover relevant changes

### DIFF
--- a/torchx/tracker/api.py
+++ b/torchx/tracker/api.py
@@ -281,7 +281,7 @@ class AppRun:
             for tracker in trackers:
                 tracker.add_source(torchx_job_id, parent_run_id, artifact_name=None)
 
-        return AppRun(id=torchx_job_id, backends=trackers_from_environ())
+        return AppRun(id=torchx_job_id, backends=trackers)
 
     def add_metadata(self, **kwargs: object) -> None:
         """Stores metadata for the current run"""
--- a/torchx/tracker/test/api_test.py
+++ b/torchx/tracker/test/api_test.py
@@ -11,8 +11,9 @@ from typing import cast, DefaultDict, Dict, Iterable, Mapping, Optional, Tuple
 from unittest import mock, TestCase
 from unittest.mock import patch
 
-import torchx.tracker as tracker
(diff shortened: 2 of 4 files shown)


torchx/tracker/api.py
```diff
@@ -281,7 +281,7 @@ def run_from_env() -> AppRun:
             for tracker in trackers:
                 tracker.add_source(torchx_job_id, parent_run_id, artifact_name=None)
 
-        return AppRun(id=torchx_job_id, backends=trackers_from_environ())
+        return AppRun(id=torchx_job_id, backends=trackers)
```
d4l3k, line 284: This should really be in a different PR -- though it's small enough

d4l3k (approved):
LGTM

d4l3k:
There's a pyre failure

torchx/tracker/api.py
kiukchung: yea... you are right, I should've done this in a different PR... just got a bit lazy at 3am... will make sure I split the PRs going forward.

torchx/tracker/test/api_test.py
```diff
@@ -119,9 +120,18 @@ def test_env_factory_test(self) -> None:
             "torchx.tracker.api.load_group",
             return_value={"tracker1": tracker_factory},
         ):
-            app_run = tracker.app_run_from_env()
+            app_run = app_run_from_env()
             self.assertEqual(app_run.id, self.run_id)
-            self.assertEqual(TestTrackerBackend, type(app_run.backends[0]))
+            self.assertEqual(1, len(app_run.backends))
```
d4l3k, line 125: likely can just change backends type to be Sized instead of Iterable

kiukchung: yep just fixed.

PR #704, opened 2023-03-15
Torchx mcad advanced networking and bug fixes

Sara-KS:
This pull request adds the following features:

- network - An option to include pod-to-pod networking in addition to standard Kubernetes networks. This option assumes the user has installed multus on their Kubernetes cluster and created an additional network configuration. Documentation updates include references for further details.
- priority_class_name - On some Kubernetes clusters, named pod priority classes may be configured. This option adds the priorityClassName field to the pods. The addition of this feature improves priority scheduling behavior on Kubernetes clusters that use MCAD priorities in combination with pod PriorityClasses.

It also includes the following bug fix:
- Kubernetes services follow the [RFC 1035 naming requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names). This bug fix removes integer values at the start of the Kubernetes object names via the cleanup_str function. 

Finally, there is a small update to the AppWrapper metadata to include the Kubernetes namespace.

Test plan:
Each item listed above includes updates to the Kubernetes-MCAD CI tests.

### DIFF
--- a/docs/source/app_best_practices.rst
+++ b/docs/source/app_best_practices.rst
@@ -105,7 +105,7 @@ model definition from a python file and then you'll load the weights and state
 dict from a ``.ckpt`` or ``.pt`` file.
 
 This is how Pytorch Lightning's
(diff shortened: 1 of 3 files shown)


d4l3k (approved):
LGTM seems very reasonable. Thanks!

PR #707, opened 2023-03-17
(torchx/tracker) Add support for MLflowTracker

kiukchung:
Adds support for an mlflow backend for the torchx.tracker. As part of this addition I'm improving a few other things:

1. Adds `torchx.distributed` module with convenience methods for using ``torch.distributed`` in a process-group agnostic manner. (Required for `on_rank0_first` context manager needed to create the mlflow once on rank0 to avoid races for ddp-style runs).
2. Adds `dist.spmd` component that sticks `TORCHX_EXPERIMENT_NAME` as an env var. The experiment name is parsed from the `--name` argument. Plan is to deprecate `dist.ddp` in favor of `dist.spmd` over time.
3. Adds two useful test-fixtures (`TestWithTmpDir` and `DistributedTestCase`). The former creates tmp directories and exposes useful methods to create temporary test files. The later exposes a method to run a function with `torchrun` to run it as a ddp script.

**NOTE**
--------------

1. I realized there is some duplication of argument parsing between `dist.spmd` and `dist.ddp`. That was intentional since I didn't want to touch `dist.ddp` as there are folks internal/external to Meta using it and I didn't want to break BC.
2. Will follow up with documentation changes.
3. Have not hooked up mlflow tracker yet since it doesn't quite work unless the job is launched via `dist.spmd`. This is because the tracker API needs a bit of a revamp to make it work generically. I'm planning to do this in a follow up PR along with some torchxconfig revamp work.

Test plan:
------------

Included unittests for changed / new features.


torchx/distributed/__init__.py
```diff
@@ -0,0 +1,261 @@
...
+            "`LOCAL_RANK` environment variable is not set. Will trivially return 0 for local_rank.\n"
+            " It is recommended to use torchrun/torchx to run your script or set the `LOCAL_RANK` manually.\n"
+            " For additional details see:\n"
+            "  1) https://pytorch.org/torchx/latest/components/distributed.html\n"
+            "  2) https://pytorch.org/docs/stable/elastic/run.html\n"
+            "=============================================================================================="
+        )
+        return 0
+    else:
+        return int(os.environ["LOCAL_RANK"])
```
d4l3k, line 52: torch distributed doesn't provide a helper, huh

torchx/components/dist.py
```diff
@@ -80,6 +86,293 @@
...
+        mounts: (for docker based runs only) mounts to mount into the worker environment/container
+                (ex. type=<bind/volume>,src=/host,dst=/job[,readonly]).
+        debug: whether to run with preset debug flags enabled
+
+    """
+
+    if env is None:
+        env = {}
+
+    argname = ArgName.parse_from(
```
d4l3k, line 351: is it possible to make this backwards compatible and just merge this into ddp? or do you want assertion behavior if it's not in the right format?

torchx/components/dist.py
```diff
@@ -80,6 +86,293 @@
...
+        if not run_name:
+            if m:  # use the last module name
+                run_name = m.rpartition(".")[2]
+            else:  # use script name w/ no extension
+                run_name = Path(safe_cast(str, script)).stem
+        return ArgName(experiment_name or None, run_name)
+
+
+@dataclass
+class ArgJ:
```
d4l3k, line 203: make private?

also I don't love this name, _StructuredJ?

torchx/components/dist.py
```diff
@@ -80,6 +86,293 @@
 """
 
 
+@dataclass
+class ArgName:
```
d4l3k, line 90: nit: maybe _StructuredName?

torchx/distributed/__init__.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+    """
+    return local_rank() == 0
+
+
+Backend = Literal["nccl", "gloo", "auto"]
+
+
+def init_pg(backend: Backend = "auto", **kwargs: Any) -> torch.device:
```
d4l3k, line 130: is this going away with 2.0?

torchx/test/fixtures.py
```diff
@@ -0,0 +1,167 @@
...
+            nproc_per_node=world_size,
+            rdzv_backend="c10d",
+            rdzv_endpoint="localhost:0",
+            max_restarts=0,
+            monitor_interval=0.01,
+            log_dir=str(self.tmpdir),
+            tee=Std.ALL,
+        )
+
+        return elastic_launch(config, entrypoint=fn)
```
d4l3k, line 167: maybe we should provide a torchx method to do this w/ pickle

torchx/tracker/mlflow.py
```diff
@@ -0,0 +1,337 @@
...
+
+from torchx.distributed import on_rank0_first
+from torchx.runner.config import get_configs
+from torchx.tracker.api import Lineage, TrackerArtifact, TrackerBase, TrackerSource
+
+log: Logger = getLogger(__name__)
+TAG_ARTIFACT_MD_PREFIX = "torchx.artifact.metadata"
+
+
+class MLflowTracker(TrackerBase):
```
d4l3k, line 27: wow this is the right capitalization

torchx/tracker/mlflow.py
```diff
@@ -0,0 +1,337 @@
...
+            if self.experiment is None:
+                experiment_id = mlflow.create_experiment(
+                    name=experiment_name,
+                    artifact_location=artifact_location,
+                )
+                self.experiment = mlflow.get_experiment(experiment_id)
+                log.info(
+                    f"Created new experiment `{experiment_name}` (id={experiment_id})"
+                )
+            else:
```
d4l3k, line 76: do we need any safety checks to avoid accidental clobbering? I suppose with the random ID it's fine

torchx/distributed/__init__.py
d4l3k, line 1: this should probably be a separate PR from the MLflow stuff

torchx/components/dist.py
```diff
@@ -80,6 +86,293 @@
...
+    return f"default-experiment/{getuser()}/{socket.getfqdn()}"
+
+
+def spmd(
+    *args: str,
+    script: Optional[str] = None,
+    m: Optional[str] = None,
+    image: str = torchx.IMAGE,
+    name: str = "/",
+    h: str = "aws_p3.2xlarge",
```
d4l3k, line 296: this shouldn't default to aws

d4l3k (commented):
exciting to see, just had some questions about long term usage

d4l3k:
cc @kurman for first OSS tracker impl

torchx/distributed/__init__.py
kiukchung: yeah... I thought there was as `dist.get_local_rank()` too... probaby should upstream this function to `dist_c10d.py`

torchx/components/dist.py
kiukchung: done. Also renamed to `StructuredNameArgument` and moved both `StructuredJArgument` and `StructuredNameArgument` to `torchx.components.structured_arg`

torchx/components/dist.py
kiukchung: renamed to `StructuredJArgument` and moved to module `torchx.components.structured_arg` so that we can reuse these higher-level arguments going forward to keep all the component arguments consistent.

torchx/components/dist.py
kiukchung: renmaed to `StructuredNameArgument`

torchx/distributed/__init__.py
kiukchung: kind of, but not exactly the way its done here. In 2.0 if you create a pg without a backend, it'll run collectives on cpu or gpu dynamically based on the tensor's device. What I want here is to chose (for all tensors) a backend cpu or gpu (gloo vs nccl) upfront based on the availability of CUDA devices.

torchx/test/fixtures.py
kiukchung: so this actually launches the function (using torch.multiprocessing) under the hood. Its not playing the "write the function into a script" trick.

torchx/tracker/mlflow.py
kiukchung: yep I was surprised myself :)

torchx/tracker/mlflow.py
kiukchung: yeah torchx job names are unique enough. The `on_rank0_first` guard seems to be enough - based on the distributed runs we've been running with this tracker for the past month.

torchx/distributed/__init__.py
kiukchung: the mlflow tracker needs `torchx.distributed.on_rank0_first` to avoid races.

torchx/components/dist.py
kiukchung: created generic named resources and defaulted this to `gpu.small`

kiukchung:
thanks for reviewing @d4l3k  addressed comments and updated the PR.

torchx/test/fixtures.py
```diff
@@ -0,0 +1,170 @@
...
```
d4l3k, line 170: oh this breaks coverage, unfortunate but not much way around it I expect?

torchx/specs/__init__.py
```diff
@@ -55,12 +58,15 @@
 def _load_named_resources() -> Dict[str, Callable[[], Resource]]:
     resource_methods = load_group("torchx.named_resources", default={})
     materialized_resources: Dict[str, Callable[[], Resource]] = {}
-    default = AWS_NAMED_RESOURCES
-    for name, resource in default.items():
+
+    for name, resource in {**GENERIC_NAMED_RESOURCES, **AWS_NAMED_RESOURCES}.items():
```
d4l3k, line 62: materialized_resources = {**GENERIC_NAMED_RESOURCES, **AWS_NAMED_RESOURCES, **resource_methods}

d4l3k (approved):
LGTM

torchx/specs/__init__.py
kiukchung: good catch fixed.

torchx/test/fixtures.py
kiukchung: hmmm we can switch to using pytest which has coverage for subprocess as well

PR #708, opened 2023-03-18
runopts: infer from TypedDict

d4l3k:
This adds a new `runopts.from_typed_dict` method and uses it to generate the runopts from the typed dict field, annotations, default parameters and docstring.

This simplifies adding new fields to schedulers since you only need to define it in a single spot.

I've migrated DockerScheduler to use the new runopts and will do the rest in a follow up PR.

Test plan:

```
pytest torchx/specs/test/api_test.py torchx/schedulers/test/api_test.py

(torchx) rice@ariel ~/D/torchx (runopts_typedict)> torchx runopts local_docker
local_docker:
    usage:
        [copy_env=COPY_ENV],[image_repo=IMAGE_REPO]

    optional arguments:
        copy_env=COPY_ENV (typing.List[str], None)
            list of glob patterns of environment variables to copy if not set in AppDef. Ex: FOO_*",
        image_repo=IMAGE_REPO (str, None)
            (remote jobs) the image repository to use when pushing patched images, must have push access. Ex: example.com/your/con
tainer

```

### DIFF
--- a/torchx/specs/api.py
+++ b/torchx/specs/api.py
@@ -6,9 +6,10 @@
 # LICENSE file in the root directory of this source tree.
 
 import copy
+import dataclasses
 import json
 import re
-from dataclasses import asdict, dataclass, field
+from dataclasses import asdict, dataclass, field, Field
 from datetime import datetime
 from enum import Enum
 from string import Template
@@ -28,7 +29,12 @@ from typing import (
     Union,
 )
 
-from torchx.util.types import to_dict
+import docstring_parser
+
+import typing_inspect
+
+from torchx.util.types import decode_optional, to_dict
+from typing_extensions import TypedDict
 
 _APP_STATUS_FORMAT_TEMPLATE = """AppStatus:
     State: ${state}
@@ -736,6 +742,11 @@ class runopts:
     def __len__(self) -> int:
         return len(self._opts)
 
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, self.__class__):
+            return self._opts == other._opts
+        return False
+
     @staticmethod
     def is_type(obj: CfgVal, tp: Type[CfgVal]) -> bool:
         """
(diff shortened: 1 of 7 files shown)


d4l3k:
cc @Sara-KS for the mcad changes -- should be a no-op, just moving image_repo opt to pull from the DockerWorkspaceMixin

torchx/specs/api.py
```diff
@@ -923,6 +934,34 @@ def __repr__(self) -> str:
...
+        for name, field_type in typed_dict.__annotations__.items():
+            default = getattr(typed_dict, name, None)
+            if isinstance(default, Field):
+                default_factory = default.default_factory
+                if default_factory is dataclasses.MISSING:
+                    default = None
+                else:
+                    # pyre-fixme[29]: not a function
+                    default = default_factory()
+            required = not (
```
kiukchung, line 952: this is going to get interesting in python 3.11 with [PEP-0655](https://peps.python.org/pep-0655/) that introduces [`typing.Required`](https://docs.python.org/3/library/typing.html#typing.Required) and [`typing.NotRequired`](https://docs.python.org/3/library/typing.html#typing.NotRequired). 

A couple of interesting questions:

1. how do we interpret a `Required[Optional[int]]`?  (disallow it? allow "required" Optional cfgs? or implicitly make this required?)
2. are there differences between declaring a cfg as `Optional[int]` versus `NotRequired[int]`? (technically the former would be in the runopt as `None` by default, and the later would simply not exist in runopt <- this concept doesn't exist in runopt yet so we'd have to default it to `None`, but then the typing breaks)

kiukchung (commented):
LGTM overall, commented on a philosophical question around the new `typing.Required` and `typing.NotRequired` types in python 3.11 and their co-usages with `Optional`

kiukchung:
Closing this long-stale experiment - the scheduler/runopts APIs have moved on since 2023. Happy to revisit the idea against current main.

PR #710, opened 2023-03-18
component_integration_tests: use compute_world instead of dummy_app

d4l3k:
This switches component_integration_tests to use an app that actually tests connectivity between nodes instead of just exiting.

Test plan:

```
scripts/component_integration_tests.py --scheduler local_docker --container_repo ""
```

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -13,7 +13,7 @@ hydra-core
 ipython
 kfp==1.8.9
 mlflow-skinny
-moto==4.1.3
+moto==4.1.6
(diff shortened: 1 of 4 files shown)


kiukchung:
nice! LGTM

kiukchung (approved):

priyaramani (approved):
Thanks for doing this! 
Looks like the integration test is failing?

d4l3k:
integration test is failing due to the changes in moto not being released yet -- rebasing now and it should pass

PR #723, opened 2023-04-05
Kubernetes mcad retry support

Sara-KS:
This PR adds retry support for the application level for Kubernetes-MCAD scheduler and updates the Kubernetes-MCAD documentation accordingly. The retry value must be 1 or higher to enable this option. This retry support applies when a job is pre-empted or the number of expected pods do not match the number of running pods. In these events, the job will be requeued with an exponentially increasing wait time between requeue checks up to the maximum number of retries specified. In the case of more than one role, the minimum of the role max retries is set as the retry value.

More details on the underlying values can be found [here](https://github.com/project-codeflare/multi-cluster-app-dispatcher/blob/main/deployment/mcad-controller/templates/deployment.yaml). 

This PR also contains a bug fix. Some Kubernetes objects are limited to only 63 characters. The update to the code includes a function to check the generated Kubernetes object names, inform the make_unique function of the number of characters available, or fails with a message to change the job name.

Test plan:
The retry feature can be tested by running an example with dist.ddp and setting the --max_retries field to 1 or higher. In the yaml, or dryrun output, users should see the additional field `schedulingSpec` with default values and the specified max_retries. Additionally, CI tests were updated to reflect the added capability. 

For the name size bug fix, a test_get_unique_id_size function was added and any CI tests including podGroup names were updated.


torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -475,6 +491,19 @@ def get_port_for_service(app: AppDef) -> str:
     return port
 
 
+def enable_retry(
+    job_spec: Dict[str, Any], appwrapper_retries: int, total_pods: int
+) -> None:
+    requeue_dict = {
+        "timeInSeconds": 300,
+        "maxTimeInSeconds": 0,
+        "growthType": "exponential",
```
d4l3k, line 500: it's cool MCAD exposes this much control

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -490,16 +519,25 @@ def app_to_resource(
...
+    MCAD supports retries at the APPLICATION level. In the case of multiple TorchX Roles,
+    the AppWrapper maximum number of retries
+    count is set to the minimum of the max_retries of the roles.
     """
 
     genericitems = []
 
-    unique_app_id = cleanup_str(make_unique(app.name))
+    unique_id_size = get_unique_id_size(app.name, len(app.roles))
+    if unique_id_size == 0:
```
d4l3k, line 530: I'd argue that 1 character is too small for sufficient uniqueness. 2 is at ~1.3k, 3 is ~46k

It might be better to truncate app.name rather than reduce ID length

d4l3k (approved):
generally seems reasonable, let me know when it's ready to merge

torchx/schedulers/kubernetes_mcad_scheduler.py
Sara-KS: Thanks for the suggestion. Updated to truncate app.name and keep at least 3 unique ID characters. Commit: 1a2ad6e90169c6ad694dca38d88610e1ecfe976c

torchx/schedulers/kubernetes_mcad_scheduler.py
```diff
@@ -527,11 +527,12 @@ def app_to_resource(
     genericitems = []
 
     unique_id_size = get_unique_id_size(app.name, len(app.roles))
-    if unique_id_size == 0:
-        raise ValueError(
-            "Name size has too many characters for some Kubernetes objects. Refer to your \
-TorchX component documentation for options to change the job name."
-        )
+    if unique_id_size == 3:
```
d4l3k, line 530: might be best to just make a new make_unique_trunc method that does this all rather than inline+get_unique_id_size.

That way you could test this logic directly

Also checking for size exactly equal to 3 seems a bit fragile

Sara-KS: Update to a consolidated function. commit: 191859d3356f8c8effac9cea2a9e2f6a39a1f728

PR #724, opened 2023-04-07
Fix build issues with torchx imports

kurman:
Summary: Fix build issues arose from difftrain changes.

Differential Revision: D44793370


priyaramani (approved):
Approving to unblock. Maybe add a TODO to fix things later.

d4l3k:
FYI we unpinned the pyre version as it wasn't being updated anymore by the Meta fbcode pyre tool. Maybe that's why it's causing these issues? We can repin if the internal tooling to periodically update it is fixed

kiukchung:
no activity, closing

PR #727, opened 2023-04-14
(torchx/tracking)(bugfix) Allow multiple ranks to properly add artifa…

kiukchung:
Summary
----------------

Fixes a bug when calling `apprun.add_artifact()` with the mlflow tracker in a distributed setting, only the artifact that is logged from the rank that won the race to first create the run. This happened b/c the ranks that lost the race and fell back to searching for the run were not setting the run_id as the "active_run" (mlflow uses a global var to store the "current active run" which is only set when you call `mlflow.start_run(run_id)`)

Misc
------

Some misc fixes in `torchx.distributed` are bundled with this PR since they were necessary for me to author a dummy app to integ test against an actual mlflow endpoint.

Test plan
-----------

Added unittest to cover this case.

Validated that artifacts from multiple ranks can indeed be logged against an actual mlflow endpoint
image

### DIFF
--- a/torchx/distributed/__init__.py
+++ b/torchx/distributed/__init__.py
@@ -16,7 +16,10 @@ from typing import Any, Iterator
 
 import torch
 import torch.distributed as dist
+from torch.distributed.distributed_c10d import _get_default_group
 from typing_extensions import Literal
+
+from torchx.util.cuda import has_cuda_devices
 
 log: logging.Logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


torchx/tracker/mlflow.py
```diff
@@ -129,10 +152,14 @@ def get_run(self, run_name: str) -> Run:
...
+                    run_name=run_name,
                 )
             elif len(search_result) == 1:
-                return search_result[0]
+                return mlflow.start_run(
+                    experiment_id=self.experiment_id,
+                    run_id=search_result[0].info.run_id,
+                )
             else:  # len(search_result) > 1
                 raise RuntimeError(
```
d4l3k, line 164: what's the risk of this race condition? should we always do a second check after calling start_run and deleting the one with the larger run_id?

torchx/distributed/__init__.py
```diff
@@ -16,8 +16,11 @@
 
 import torch
 import torch.distributed as dist
+from torch.distributed.distributed_c10d import _get_default_group
 from typing_extensions import Literal
 
+from torchx.util.cuda import has_cuda_devices
+
 log: logging.Logger = logging.getLogger(__name__)
```
d4l3k, line 26: For local_rank should we be returning os.environ["LOCAL_RANK"] if dist isn't initialized to match the new behavior of the other methods? Not sure if that was an oversight. 

Line 40 since I can't directly comment on that via GitHub ui

d4l3k (approved):
seems reasonable

torchx/tracker/mlflow.py
kiukchung: the race is actually handled in the ctor of `MLflowTracker` where we wrap the create-or-query experiment/run `with on_rank0_first():`. The issue that this PR is solving is that when the non-zero ranks query the run, we were not **setting** the active run with the queried run_id which means that we never end up logging artifacts, metrics, etc against the job's run_id.

torchx/distributed/__init__.py
kiukchung: yeah good point. Fixed `local_rank()` to return `LOCAL_RANK` if set, otherwise 0.

Issue #709, opened 2023-03-18
deprecate Python 3.7 support


d4l3k:
## Description

PyTorch 2.0 has deprecated Python 3.7 support. We should drop it as well since it's painful to support 3.7 with modern types.

https://pytorch.org/get-started/pytorch-2.0/

Issue #726, opened 2023-04-14
Codecov pip install fails in TorchX GitHub actions


Sara-KS:
## 🐛 Bug

Installing Codecov via pip install is no longer a supported option as mentioned in the [Codecov blog post from April 12th.](https://about.codecov.io/blog/message-regarding-the-pypi-package/)  This change is currently causing the python unit test GitHub action to fail. The recommended solution from Codecov is to migrate to using the Codecov Github Action. 

Module (check all that applies):
 * [ ] `torchx.spec`
 * [ ] `torchx.component`
 * [ ] `torchx.apps`
 * [ ] `torchx.runtime`
 * [ ] `torchx.cli`
 * [ ] `torchx.schedulers`
 * [ ] `torchx.pipelines`
 * [ ] `torchx.aws`
 * [ ] `torchx.examples`
 * [ x] `other`

## To Reproduce

1. Trigger running the TorchX Github action: .github/workflows/python-unittests.yaml 
2 . pip install coverage codecov line fails

Error message:
```
ERROR: Could not find a version that satisfies the requirement codecov (from versions: none)
ERROR: No matching distribution found for codecov
```

## Environment

 - torchx version (e.g. 0.1.0rc1):
 - Python version: All
 - OS (e.g., Linux):
 - How you installed torchx (`conda`, `pip`, source, `docker`):
 - Docker image and tag (if using docker):
 - Git commit (if installed from source):
 - Execution environment (on-prem, AWS, GCP, Azure etc):
 - Any other relevant information: GitHub Actions

kiukchung:
Thanks for reporting! I've fixed this as part of my PR: https://github.com/pytorch/torchx/pull/727

PR #728, opened 2023-04-16
(torchx/python) Drop support for python 3.7

kiukchung:
Summary
-----------
Following torch-2.0, drops support for python 3.7.
Resolves: https://github.com/pytorch/torchx/issues/709

FWIW python 3.7 currently doesn't work due to torchx requiring `mlflow-skinny>=2` but `mlflow-skinny` v2 not being available for python 3.7.

Test plan:
-----------
CI

### DIFF
--- a/.github/workflows/python-unittests.yaml
+++ b/.github/workflows/python-unittests.yaml
@@ -10,7 +10,7 @@ jobs:
   unittest:
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10']
+        python-version: [3.8, 3.9, '3.10']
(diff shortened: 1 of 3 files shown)


d4l3k (approved):
LGTM -- looks like there's an issue with codecov? seems unrelated but would be good to fix

kiukchung:
Python Unittests in CI will fail due to https://github.com/pytorch/torchx/issues/726, which I resolved in https://github.com/pytorch/torchx/pull/727 but not sure how to do stacked diffs on github. So I'll merge 727 first, then merge this manually.

PR #731, opened 2023-04-19
(torchx/aws_batch)(bugfix) Correctly query the CW logStreamName for n…

kiukchung:
Summary
---------------

Fixes a bug in the `log_iter()` method of `aws_batch_scheduler` where the `logStreamName` was always being set as the `logStreamName` of node 0. So the logs of node 0 were being printed out for all nodes. 

Example: (2 node, 2 worker job - `-j 2x2`)

```
$ torchx log aws_batch://torchx/MY_QUEUE/MY_JOB

ddp/0 [0] INFO: logging from node 0 rank 0
ddp/0 [1] INFO: logging from node 0 rank 1
ddp/1 [0] INFO: logging from node 0 rank 0  <-- node 0 log showing up as node 1 log!
ddp/1 [1]  INFO: logging from node 0 rank 1 <-- node 0 log showing up as node 1 log!
``` 

Test plan:
-------------

Modified job description mock in unittest to cover for this case.

Ran a multi-node job and validated that `torchx log` is indeed correctly prints logs for all nodes (rather than node0's for every node).

### DIFF
--- a/torchx/schedulers/test/aws_batch_scheduler_test.py
+++ b/torchx/schedulers/test/aws_batch_scheduler_test.py
@@ -20,6 +20,7 @@ from torchx.schedulers.aws_batch_scheduler import (
     AWSBatchOpts,
     AWSBatchScheduler,
     create_scheduler,
+    ENV_TORCHX_ROLE_NAME,
     resource_from_resource_requirements,
     resource_requirements_from_resource,
     to_millis_since_epoch,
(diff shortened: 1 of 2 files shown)


torchx/schedulers/test/aws_batch_scheduler_test.py
```diff
@@ -342,40 +343,91 @@ def _mock_scheduler_running_job(self) -> AWSBatchScheduler:
...
+                        },
+                        "tags": {
+                            "torchx.pytorch.org/version": "0.3.0dev0",
+                            "torchx.pytorch.org/app-name": "fairseq-train",
+                        },
+                        "platformCapabilities": [],
+                    }
+                ]
+            },
+        ]
```
d4l3k, line 430: can we add test to verify the log streams are actually queried?

d4l3k (approved):
LGTM, nice catch

kurman (approved):

torchx/schedulers/test/aws_batch_scheduler_test.py
kiukchung: the test actually does this (but its all mocked so it doesn't actually check logstream names (e.g. the mock CW client just returns whatever it was told to return). I need to make another change to the batch scheduler (for some reason the list API doesn't return all jobs) will revamp the tests to use moto.

PR #732, opened 2023-04-22
(torchx/local_scheduler) Cleanup auto_set_cuda_visible_devices logic …

kiukchung:
Summary
--------------

1. Cleans up the logic to auto-set CUDA_VISIBLE_DEVICES in local_scheduler.
2. Adds more testing around it
3. Fixes a small bug in `dist.ddp` where the port mapping on the ddp role was hard-coded to 29500 rather than using `rdzv_port`.

Test plan
------------
```
$ pytest
```

### DIFF
--- a/torchx/schedulers/local_scheduler.py
+++ b/torchx/schedulers/local_scheduler.py
@@ -45,15 +45,15 @@ from torchx.specs.api import AppDef, AppState, is_terminal, macros, NONE, Role, 
 from torchx.util.types import none_throws
 from typing_extensions import TypedDict
 
-
 log: logging.Logger = logging.getLogger(__name__)
 
 STDOUT_LOG = "stdout.log"
 STDERR_LOG = "stderr.log"
 COMBINED_LOG = "combined.log"
 
-
 NA: str = "<N/A>"
+
+ENV_CUDA_VISIBLE_DEVICES = "CUDA_VISIBLE_DEVICES"
 
 
 class SignalException(Exception):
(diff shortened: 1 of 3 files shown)


torchx/schedulers/local_scheduler.py
```diff
@@ -763,7 +763,11 @@ def _submit_dryrun(
             request, lambda p: pprint.pformat(asdict(p), indent=2, width=80)
         )
 
-    def _get_gpu_device_count(self) -> int:
+    def _device_count(self) -> int:
+        # this method deliberately does not use ``torch.cuda.device_count()``
+        # to avoid taking a dependency on pytorch
+        # this make sit possible to avoid a BUCK dependency (internally at Meta)
+        # on //caffe2:torch which slows down builds of //torchx:* rules
```
d4l3k, line 770: I stumbled upon pyNVML library https://pypi.org/project/nvidia-ml-py/ doesn't quite solve the buck issue but would be lighterweight than torch perhaps? Just something I was thinking about, this is fine

torchx/schedulers/local_scheduler.py
```diff
@@ -763,7 +763,11 @@ def _submit_dryrun(
             request, lambda p: pprint.pformat(asdict(p), indent=2, width=80)
         )
 
-    def _get_gpu_device_count(self) -> int:
+    def _device_count(self) -> int:
```
d4l3k, line 766: nit: rename to _nvidia_device_count?

d4l3k (approved):
LGTM

torchx/schedulers/local_scheduler.py
kiukchung: yeah it'll be nice to use that python-API, I'm not sure if `nvidia-ml-py` has been imported as a 3rd party lib in BUCK (I'm sure it has but I can't say 100% for sure) and how it'll make things better/worse for torchx in terms of BUCK deps/buildtime etc. Perhaps this is something @kurman or @priyaramani can help us with.

torchx/schedulers/local_scheduler.py
kiukchung: done, renamed to `_cuda_device_count`

PR #736, opened 2023-07-05
Add option to specify a subset of CUDA devices for the job to run on

miqueljubert:
Summary:
Add a new parameters, cuda_visible_devices_subset, which contains a list of GPU indices. If set, auto_set_cuda_visible_devices will only use indices from this list when distributing the indices.

This allows masking out some GPUs, useful in scenarios like hosts shared between multiple users, where the first GPUs will be often in use by default processes, and hosts which have different types of GPUs available, and it is desired to only use a subset of those.

Differential Revision: D47208267

### DIFF
--- a/torchx/schedulers/local_scheduler.py
+++ b/torchx/schedulers/local_scheduler.py
@@ -168,6 +168,7 @@ class LocalOpts(TypedDict, total=False):
     log_dir: str
     prepend_cwd: Optional[bool]
     auto_set_cuda_visible_devices: Optional[bool]
+    auto_set_cuda_visible_devices_ids: List[str]
 
 
 class LocalDirectoryImageProvider(ImageProvider):
@@ -595,6 +596,14 @@ class LocalScheduler(Scheduler[LocalOpts]):
             help="sets the `CUDA_AVAILABLE_DEVICES` for roles that request GPU resources."
             " Each role replica will be assigned one GPU. Does nothing if the device count is less than replicas.",
         )
+        opts.add(
+            "auto_set_cuda_visible_devices_ids",
+            type_=List[str],
+            default=[],
+            help="when auto_set_cuda_visible_devices is set, the ; separated list of GPUs to use "
+            "during auto-setting CUDA_VISIBLE_DEVICES. If not set, defaults to all devices "
+            "available on the machine.",
+        )
         return opts
 
     def _validate(self, app: AppDef, scheduler: str) -> None:
@@ -763,7 +772,7 @@ class LocalScheduler(Scheduler[LocalOpts]):
             request, lambda p: pprint.pformat(asdict(p), indent=2, width=80)
         )
 
-    def _cuda_device_count(self) -> int:
+    def _cuda_device_count(self) -> Optional[int]:
         # this method deliberately does not use ``torch.cuda.device_count()``
         # to avoid taking a dependency on pytorch
         # this makes it possible to avoid a BUCK dependency (internally at Meta)
@@ -780,6 +789,11 @@ class LocalScheduler(Scheduler[LocalOpts]):
         except subprocess.CalledProcessError as e:
             log.exception(f"Got exception while listing GPUs: {e.stderr}")
             return 0
+        except FileNotFoundError as e:
+            log.error(
+                "nvidia-smi is not installed, could not automatically fetch available GPUs"
+            )
+            return None
 
     def auto_set_CUDA_VISIBLE_DEVICES(
         self,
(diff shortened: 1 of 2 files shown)


d4l3k (commented):
this seems pretty reasonable to me though @kiukchung  has better context on this logic

torchx/schedulers/local_scheduler.py
```diff
@@ -168,6 +168,7 @@ class LocalOpts(TypedDict, total=False):
     log_dir: str
     prepend_cwd: Optional[bool]
     auto_set_cuda_visible_devices: Optional[bool]
+    auto_set_cuda_visible_devices_ids: List[str]
```
kurman, line 171: Should the list be an optional? I think we need an explicit optional semantics instead of empty list.

kiukchung:
I think the logic here can be simplified quite a bit if we do the following:

1. If `auto_set_cuda_visible_devices_ids` is passed, then trust the user-input and simply use that list of cuda_devices to auto-set `CUDA_VISIBLE_DEVICES` based on the number of ddp procs to run (e.g. the product of `-j #x#`). No need to validate it against `nvidia-smi`.

2. Make this option type `List[int]` instead of `List[str]`. This way you get "is_number" validation for free.

I'm all for input validation, but this is one of the cases where the code is getting hard to read/maintain.

kurman:
> 1. No need to validate it against `nvidia-smi`.

+1 on not relying on binary on the PATH. In addition to that jobs typically run on homogenous hardware.

torchx/schedulers/local_scheduler.py
miqueljubert: TorchX right now does not understand optional scheduler arguments. 

```
Cache hits: 100%. Commands: 85 (cached: 85, remote: 0, local: 0)
Traceback (most recent call last):
  File "<string>", line 51, in <module>
  File "<string>", line 37, in __run
  File "/usr/local/fbcode/platform010/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/fbcode/platform010/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/data/users/jmiquel/fbsource/buck-out/v2/gen/fbcode/da4de3c780a17bfa/torchx/cli/__torchx__/torchx#link-tree/torchx/cli/main.py", line 120, in <module>
    main()
  File "/data/users/jmiquel/fbsource/buck-out/v2/gen/fbcode/da4de3c780a17bfa/torchx/cli/__torchx__/torchx#link-tree/torchx/cli/main.py", line 116, in main
    run_main(get_sub_cmds(), argv)
  File "/data/users/jmiquel/fbsource/buck-out/v2/gen/fbcode/da4de3c780a17bfa/torchx/cli/__torchx__/torchx#link-tree/torchx/cli/main.py", line 112, in run_main
    args.func(args)
  File "/data/users/jmiquel/fbsource/buck-out/v2/gen/fbcode/da4de3c780a17bfa/torchx/cli/__torchx__/torchx#link-tree/torchx/cli/cmd_run.py", line 248, in run
    self._run(runner, args)
  File "/data/users/jmiquel/fbsource/buck-out/v2/gen/fbcode/da4de3c780a17bfa/torchx/cli/__torchx__/torchx#link-tree/torchx/cli/cmd_run.py", line 184, in _run
    cfg = scheduler_opts.cfg_from_str(args.scheduler_args)
  File "/data/users/jmiquel/fbsource/buck-out/v2/gen/fbcode/da4de3c780a17bfa/torchx/cli/__torchx__/torchx#link-tree/torchx/specs/api.py", line 866, in cfg_from_str
    cfg[key] = _cast_to_type(val, runopt_.opt_type)
  File "/data/users/jmiquel/fbsource/buck-out/v2/gen/fbcode/da4de3c780a17bfa/torchx/cli/__torchx__/torchx#link-tree/torchx/specs/api.py", line 860, in _cast_to_type
    return opt_type(value)
  File "/usr/local/fbcode/platform010/lib/python3.10/typing.py", line 957, in __call__
    result = self.__origin__(*args, **kwargs)
  File "/usr/local/fbcode/platform010/lib/python3.10/typing.py", line 387, in __call__
    raise TypeError(f"Cannot instantiate {self!r}")
TypeError: Cannot instantiate typing.Union
```

I will look at whether it can be cleanly added to the type instantiation.

miqueljubert:
I will push another diff with some changes to the torchx/specs/api.py. Otherwise neither Optional[List[str]] or Optional[List[int]] are supported CfgNode types, as per api.py

kiukchung:
is this PR still relevant?

kiukchung:
closing as there is no activity. Feel free to reopen!

PR #739, opened 2023-07-21
make rdzv_backend configurable

MichaelClifford:
PR Summary:
This PR allows the user to configure the `rdzv_backend` they want to use with `ddp` and sets the default to "c10d". This also removes the previously hard-coded  c10d rdzv_backend. 

Test plan:

### DIFF
--- a/torchx/components/dist.py
+++ b/torchx/components/dist.py
@@ -171,6 +171,7 @@ def ddp(
     env: Optional[Dict[str, str]] = None,
     max_retries: int = 0,
     rdzv_port: int = 29500,
+    rdzv_backend: str = "c10d",
     mounts: Optional[List[str]] = None,
     debug: bool = False,
 ) -> specs.AppDef:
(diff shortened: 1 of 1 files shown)


kiukchung (approved):
Thanks for the PR, LGTM. 

Two small notes:

1. Certain `rdzv_backends` require additional rendezvous configs to be passed as `rdzv_conf KEY=VAL,KEY=VAL,...`, so you might end up having to expose `rdzv_conf` as well.
2. The `spmd` component (in the same file) calls `ddp` and has a much cleaner argument list. So depending on what you are after, you might want to switch over to `spmd` and plumb `rdzv` argument through to that component as well.

Both are non-blocking  comments. Approving.

MichaelClifford:
@kiukchung thanks for the review. And happy to add those noted changes in an additional PR. 

After the approval, is there anything else I need to do before the PR can be mereged?

d4l3k:
Code looks good -- there's some CI failures in trunk which are unrelated to this (pyyaml regression + disk space issue on the underlying runners)

PR #745, opened 2023-07-26
ci: use larger instances to avoid docker disk space issues

d4l3k:
Resolves issues related to docker build size.

Test plan:

CI


d4l3k:
@kiukchung CI only changes, merging

PR #751, opened 2023-08-07
add 'node_rank' to torchrun cmd

MichaelClifford:
This PR adds the `node_rank` parameter to the torchrun command in `dist.py` and sets it to `macros.replica_id` for each node.  

This parameter appears to be required when running torchx with the ray scheduler across multiple nodes. 

Test plan:

### DIFF
--- a/torchx/components/dist.py
+++ b/torchx/components/dist.py
@@ -261,6 +261,8 @@ def ddp(
         nnodes_rep,
         "--nproc_per_node",
         str(nproc_per_node),
+        "--node_rank",
+        f"{macros.replica_id}",
         "--tee",
         "3",
         "--role",
--- a/torchx/components/dist.py
+++ b/torchx/components/dist.py
@@ -261,6 +261,7 @@ def ddp(
         nnodes_rep,
         "--nproc_per_node",
         str(nproc_per_node),
+        # node_rank is only used when rdzv_backend is 'static'
         "--node_rank",
         f"{macros.replica_id}",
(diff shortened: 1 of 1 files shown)


d4l3k:
Can you elaborate why node_rank is required? torch elastic will automatically assign ranks through the c10d tcpstore based rendezvous. Are nodes failing to connect?

d4l3k:
https://github.com/pytorch/pytorch/blob/3a300ed84efa8e458df02ee4f92532667cefbe29/torch/distributed/run.py#L718-L719 only uses node_rank for static rdzv_backend. Are you using static or c10d?

MichaelClifford:
Yeah, that is it exactly. We are using static rdzv_backend. 

Would it be better to only enable "--node_rank" option when the backend is static?

torchx/components/dist.py
```diff
@@ -261,6 +261,8 @@ def ddp(
         nnodes_rep,
         "--nproc_per_node",
         str(nproc_per_node),
+        "--node_rank",
+        f"{macros.replica_id}",
```
d4l3k, line 265: can you add a comment that this only applies when rendezvous is static?

d4l3k (approved):
generally seems reasonable -- shouldn't be an issue to specify it if it's not used

torchx/components/dist.py
MichaelClifford: Thanks @d4l3k. Done.

PR #755, opened 2023-08-16
Resolve run_opts before passing it to the workspace

manav-a:
Summary:
We were resolving the default values for runopts during the dryrun call on the scheduler. This made the ops passed to the workspace not have the defaults correctly populated for workspace opts. This change also resolves the runopts in the runner dryrun and scheduler submit apis.

Haven't removed the runopts resolution in scheduler dry run here as a lot of other tests broke with it and it seems reasonable to also have runopts resolved for just the scheduler dryrun. The double resolving of runopts for the runner dryrun and scheduler submit cases shouldnt cause any meaningful differences.

There is a separate question on whether workspace should also be built during scheduler dryrun but that can be a follow up change.

Differential Revision: D48395915

### DIFF
--- a/torchx/runner/api.py
+++ b/torchx/runner/api.py
@@ -357,7 +357,7 @@ class Runner:
         cfg = cfg or dict()
         with log_event("dryrun", scheduler, runcfg=json.dumps(cfg) if cfg else None):
             sched = self._scheduler(scheduler)
-
+            resolved_cfg = sched.run_opts().resolve(cfg)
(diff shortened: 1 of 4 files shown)


kurman (commented):
Config resolution can be eager, we probably should move to that direction.

kurman:
Created issues https://github.com/pytorch/torchx/issues/757 and https://github.com/pytorch/torchx/issues/758, those have been broken.

kurman (approved):

PR #756, opened 2023-08-16
Fix argument parsing bug

kunalb:
Summary: Ran into an error when I had a component function with an optional bool argument that was being set to None; realized that the unit tests had the function defined but didn't actually use it yet.

Differential Revision: D48384586


kiukchung:
LGTM thanks for the fix!

kiukchung (approved):
Stamping pending CI fix.

kurman (approved):
lgtm

PR #761, opened 2023-08-17
add node_rank to torchrun cmd if rdzv_backend is 'static'

MichaelClifford:
#751 
#759 

This PR adds the torchrun parameter `node_rank` in the cmd creation of `ddp()`. This parameter is needed if a user wants to use `static` rdzv_backend instead of `c10d`. 

The additional parameter is only added when the rdzv_backend is set to `static`.  For more details on why it has been implemented this way please see #759   

Test plan:

### DIFF
--- a/torchx/components/dist.py
+++ b/torchx/components/dist.py
@@ -266,6 +266,10 @@ def ddp(
         "--role",
         "",
     ]
+    # TODO 'node_rank' is made optional as it currently does not work with the AWS Batch scheduler.
+    # node_rank is only used when rdzv_backend is 'static'
+    if rdzv_backend == "static":
+        cmd += ["--node_rank", f"{macros.replica_id}"]
     if script is not None:
         cmd += [script]
     elif m is not None:
--- a/torchx/components/test/dist_test.py
+++ b/torchx/components/test/dist_test.py
@@ -38,6 +38,12 @@ class DDPTest(ComponentTestCase):
         for k, v in _TORCH_DEBUG_FLAGS.items():
             self.assertEqual(env[k], v)
 
+    def test_ddp_rdzv_backend_static(self) -> None:
+        app = ddp(script="foo.py", rdzv_backend="static")
+        cmd = app.roles[0].args[1]
+        self.assertTrue("--rdzv_backend static" in cmd)
+        self.assertTrue("--node_rank" in cmd)
+
 
 class SpmdTest(ComponentTestCase):
     def test_validate_spmd(self) -> None:


torchx/components/dist.py
```diff
@@ -266,6 +266,10 @@ def ddp(
         "--role",
         "",
     ]
+    # TODO 'node_rank' is made optional as it currently does not work with the AWS Batch scheduler.
+    # node_rank is only used when rdzv_backend is 'static'
+    if rdzv_backend == "static":
+        cmd += ["--node_rank", f"{macros.replica_id}"]
```
d4l3k, line 272: can you add a unit test for rdzv_backend=="static"

kiukchung: Is this change specifically for the AWS Batch use-case? And if so, is the intention to make it easier for users to find node 0's log stream by ensuring that the AWS Batch assigned node0 always runs ddp's rank0?

If so, using `static` rdzv_backend on AWS Batch won't work as intended because the `static` rdzv_backend was added to ensure backwards compatibility with `torch.distributed.launch` in torch versions less than 1.9. What the `static` rdzv_backend does NOT do is to "wait" for full rank before forking the actual ddp processes. Since AWS Batch is not a true gang scheduler, the way AWS Batch creates MNP (multi-node-parallel) jobs, is that it first provisions `node0`, then starts adding additional nodes one-by-one until the job gets the requested number of nodes. This process can take up to 45minutes (depending on the availability of the instance types, size of the docker image, etc) and therefore it forces you to set the process group timeout when calling `dist.init_process_group()` to something larger than 45min. Since the default pg is also used to timeout collective calls, GPU/node failures now take up to 45minutes to detect which is probably what you don't want.

A better solution is to just pin rank0 on AWS Batch's node0 as done in this sample [PR](https://github.com/pytorch/torchx/pull/750/files#diff-a154331bf57c74ef6656f75b9ff6c13bcce6c90f720e54c31f4eb5b3d55fc47bR198-R205) and still use `rdzv_backend="c10d"`. This however has the drawback that ONLY rank0 is pinned to node0. But in most cases this is what folks want.

d4l3k: This is specifically not for AWS Batch since this broke on the AWS Batch scheduler due to the macro expansion not being correct. I believe this is being used on the Ray scheduler

This was originally reverted in https://github.com/pytorch/torchx/pull/759

MichaelClifford: Thanks @d4l3k added a unit test.

d4l3k (approved):

Issue #752, opened 2023-08-14
Support headers and request verification options for Ray Client


KPostOffice:
## Description
The Ray SDK has options for specifying headers as well as a flag for HTTPS verification that are not exposed in anyway within torchx when submitting jobs.

## Motivation/Background
We are looking to put our Ray instance behind an OAuth server. This means that communicating with it will require setting an authorization header as well as specifying the self signed cert that we create for it.

## Detailed Proposal
Expose these options to torchx users.

## Alternatives
We will create a custom fork of torchx which provides these options.

## Additional context/links
https://github.com/ray-project/ray/blob/5435807cb429b89bc7bb3f61464131d41f60460f/dashboard/modules/dashboard_sdk.py#L206-207

kiukchung:
As @d4l3k mentioned on the torchx slack channel, there's a couple of approaches (with varying degrees of difficulty vs clean-ness trade-off). I'll start with the easiest (albeit hackiest):

1. **Leverage environment variables**. If the Ray client reads http headers from env vars, then you can create a custom component and hard-code the default auth headers for your organization. This requires no code change in torchx. If the Ray client doesn't support  setting http headers via env vars, then you would have to edit [`torchx.schedulers.RayScheduler`](https://github.com/pytorch/torchx/blob/main/torchx/schedulers/ray_scheduler.py#L115) to read the headers off of env vars and pass them when creating the Ray client.
2. **Add optional constructor args to `RayScheduler`**. Similar to `client` and `docker_client` optional args to the [`KubernetesScheduler`](https://github.com/pytorch/torchx/blob/main/torchx/schedulers/kubernetes_scheduler.py#L540-L541). If you are using torchx programmatically then you'd create the `RayScheduler` with the newly added extra args. To make this work from the torchx CLI as well you'd have to plumb it through the [`get_runner()`](https://github.com/pytorch/torchx/blob/3a253be910cd1029bb00bb4ed97a5c68cbd4ff10/torchx/runner/api.py#L682) method and the call-sites of `get_runner()` in the `cmd_<API>.py` (see [link](https://github.com/search?q=repo%3Apytorch%2Ftorchx+get_runner%28+path%3A%2F%5Etorchx%5C%2Fcli%5C%2F%2F&type=code)) files.
[...]

PR #762, opened 2023-08-21
add option for programatically defining ray job client

KPostOffice:
solves: #752 

I've added an optional `__init__` arg which allows the user to specify their own `RayJobSubmission` client

Test plan:
- [ ] add unit tests
- [ ] check and make sure that it solves intended use case (OAuth proxy protected dashboard)

Documentation:
- [ ] add docs for new option

### DIFF
--- a/torchx/schedulers/ray_scheduler.py
+++ b/torchx/schedulers/ray_scheduler.py
@@ -14,7 +14,9 @@ import time
 from dataclasses import dataclass, field
 from datetime import datetime
 from shutil import copy2, rmtree
-from typing import Any, cast, Dict, Iterable, List, Optional, Tuple  # noqa
+from typing import Any, cast, Dict, Final, Iterable, List, Optional, Tuple  # noqa
+
+import urllib3
 
 from torchx.schedulers.api import (
     AppDryRunInfo,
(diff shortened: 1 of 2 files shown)


torchx/schedulers/ray_scheduler.py
```diff
@@ -148,8 +151,19 @@ class RayScheduler(TmpDirWorkspaceMixin, Scheduler[RayOpts]):
 
         """
 
-        def __init__(self, session_name: str) -> None:
+        def __init__(self, session_name: str, ray_client: Optional[JobSubmissionClient] = None) -> None:
             super().__init__("ray", session_name)
+            self._ray_client = ray_client
+
+        def ray_client(self, job_submission_netloc):
```
d4l3k, line 158: can we make this private? i.e. `_get_ray_client()` to avoid naming conflicts?

torchx/schedulers/ray_scheduler.py
```diff
@@ -407,12 +419,12 @@ def list(self) -> List[ListAppResponse]:
                     "RAY_ADDRESS env variable is expected to be set to list jobs on ray scheduler."
```
d4l3k, line 419: should we merge this RAY_ADDRESS logic into `ray_client()`?

torchx/schedulers/ray_scheduler.py
```diff
@@ -407,12 +419,12 @@ def list(self) -> List[ListAppResponse]:
...
                     " See https://docs.ray.io/en/latest/cluster/jobs-package-ref.html#job-submission-sdk for more info"
                 )
-            client = JobSubmissionClient(address)
+            client = self._ray_client if self._ray_client else JobSubmissionClient(address)
```
d4l3k, line 422: can we fold this logic into ray_client?

d4l3k (commented):
Code seems reasonable -- can we add an in some tests for this custom client? I think we can use magic mock to plug in this logic and test the ray_client netloc code

torchx/schedulers/test/ray_scheduler_test.py
```diff
@@ -321,10 +326,19 @@ def test_list_throws_without_address(self) -> None:
...
                 self._scheduler.list()
 
+        def test_list_doesnt_throw_with_client(self) -> None:
+            ray_client = JobSubmissionClient(address="https://test.com")
+            ray_client.list_jobs = MagicMock(return_value=[])
+            _scheduler_with_client = RayScheduler("client_session", ray_client)
+            try:
+                _scheduler_with_client.list()
+            except:
+                self.fail("Unexpected when listing jobs.")
```
d4l3k, line 340: we can just let it throw and add a comment that we're testing for success

d4l3k (approved):
LGTM thanks for the cleanups! Seems like there's some CI failures

PR #764, opened 2023-08-31
Make --log wait automatically in run

kunalb:
Summary: I generally used to find it confusing and assume that `--log` was broken before realizing I always need to specify `--wait` to make it work -- I think it would be better to have `--log` automatically wait for the job to finish.

Differential Revision: D48874784

### DIFF
--- a/torchx/cli/cmd_log.py
+++ b/torchx/cli/cmd_log.py
@@ -101,12 +101,14 @@ def get_logs(
     if len(path) == 4:
         replica_ids = [(role_name, int(id)) for id in path[3].split(",") if id]
     else:
-        for i in range(10):
+        display_waiting = True
+        while True:
             status = runner.status(app_handle)
             if status and is_started(status.state):
                 break
-            if i == 0:
-                logger.info("Waiting for app to start before logging...")
+            elif display_waiting:
+                logger.info("Waiting for app state response before fetching logs...")
+                display_waiting = False
             time.sleep(1)
 
         app = none_throws(runner.describe(app_handle))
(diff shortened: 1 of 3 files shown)


manav-a:
Hmm do you know why this PR also has the changes in #763

torchx/cli/cmd_log.py
```diff
@@ -101,12 +101,14 @@ def get_logs(
...
-        for i in range(10):
+        display_waiting = True
+        while True:
             status = runner.status(app_handle)
             if status and is_started(status.state):
                 break
-            if i == 0:
-                logger.info("Waiting for app to start before logging...")
+            elif display_waiting:
+                logger.info("Waiting for app state response before fetching logs...")
```
kiukchung, line 110: since this is in the CLI code, rather than printing multiple log lines with the same message, why not print (to stdout) with a carriage-return (`\r`)?

torchx/cli/cmd_run.py
```diff
@@ -223,7 +223,7 @@ def _run(self, runner: Runner, args: argparse.Namespace) -> None:
                     app_status = runner.status(app_handle)
                     if app_status:
                         logger.info(app_status.format())
-                    if args.wait:
+                    if args.wait or args.log:
```
kiukchung, line 226: add test for this branch?

kiukchung (commented):
thanks for fixing

torchx/cli/cmd_log.py
kunalb: It should only display once -- display_waiting is immediately set to false. (I lost the iteration index while changing to the while loop)

torchx/cli/cmd_run.py
kunalb: ack!

kiukchung (approved):
LGTM thanks!

PR #767, opened 2023-09-15
add options to create_scheduler so that the get_runner method is full…

KPostOffice:
…y configurable

I also added a note to each schedulers __init__ method to help with maintainablility

I changed the create scheduler function for the local scheduler to specify the key word args that it takes rather than getting them from the `**kwargs` dict.

Test plan:
- [ ] add unit tests

### DIFF
--- a/torchx/schedulers/ray_scheduler.py
+++ b/torchx/schedulers/ray_scheduler.py
@@ -153,6 +153,7 @@ if _has_ray:
         def __init__(
             self, session_name: str, ray_client: Optional[JobSubmissionClient] = None
         ) -> None:
+            # NOTE: make sure any new init options are supported in create_scheduler(...)
             super().__init__("ray", session_name)
 
             # w/o Final None check in _get_ray_client does not work as it pyre assumes mutability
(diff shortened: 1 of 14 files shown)


torchx/schedulers/ray_scheduler.py
```diff
@@ -441,10 +442,12 @@ def list(self) -> List[ListAppResponse]:
...
+def create_scheduler(
+    session_name: str, ray_client: Optional[JobSubmissionClient] = None, **kwargs: Any
+) -> "RayScheduler":
     if not has_ray():  # pragma: no cover
         raise ModuleNotFoundError(
             "Ray is not installed in the current Python environment."
         )
 
-    return RayScheduler(session_name=session_name)
+    return RayScheduler(session_name=session_name, ray_client=ray_client)
```
d4l3k, line 453: would be nice to add some unit tests for this

d4l3k (approved):
LGTM -- thanks! A unit test would be nice but not critical and fixing lint

KPostOffice:
@d4l3k I've added those unit tests. Feel free to merge if you think the PR is good to go

d4l3k (approved):
LGTM thanks for adding!

Issue #744, opened 2023-07-25
Python 3.11 compatibility


saitcakmak:
Python 3.11 does not support mutable defaults in dataclasses. [`Role`](https://github.com/pytorch/torchx/blob/main/torchx/specs/api.py#L336) uses a mutable default for `resource`, which leads to a `ValueError` when importing torchx in Python 3.11. This prevents downstream packages from supporting Python 3.11. Example failure: https://github.com/facebook/Ax/actions/runs/5662890368/job/15343645668?pr=1744

Possible fixes: 
- Make `Resource` a frozen dataclass, which would make it immutable, likely fixing the issue.
- Set the default value to `MISSING` and set the current default in a `__post_init__` method.

bryantbiggs:
related https://github.com/pytorch/pytorch/pull/91795

henrylhtsang:
Any update on this?

kiukchung:
https://github.com/pytorch/torchx/pull/768 oughta fix it.

saitcakmak:
Hi @kiukchung. I am still getting an error with Py3.11.
```
(py311) saitcakmak@saitcakmak-mbp ~ % pip install torchx-nightly==2023.9.19
...
(py311) saitcakmak@saitcakmak-mbp ~ % python
Python 3.11.4 (main, Jul  5 2023, 08:40:20) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import torchx
>>> from torchx.specs.api import Role
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/anaconda3/envs/py311/lib/python3.11/site-packages/torchx/specs/__init__.py", line 16, in <module>
    from torchx.specs.named_resources_aws import NAMED_RESOURCES as AWS_NAMED_RESOURCES
  File "/opt/anaconda3/envs/py311/lib/python3.11/site-packages/torchx/specs/named_resources_aws.py", line 34, in <module>
    from torchx.specs.api import Resource
  File "/opt/anaconda3/envs/py311/lib/python3.11/site-packages/torchx/specs/api.py", line 272, in <module>
    @dataclass
     ^^^^^^^^^
  File "/opt/anaconda3/envs/py311/lib/python3.11/dataclasses.py", line 1230, in dataclass
    return wrap(cls)
           ^^^^^^^^^
  File "/opt/anaconda3/envs/py311/lib/python3.11/dataclasses.py", line 1220, in wrap
[...]

PR #768, opened 2023-09-15
[torchx/specs] Use default_factory for the default value of Role.reso…

kiukchung:
Also adds 3.11 to the python-unittests CI action matrix.

Fixes: https://github.com/pytorch/torchx/issues/744


d4l3k (approved):
LGTM

PR #769, opened 2023-09-18
Migrate Batch integration tests to use AWS resources in different region

kurman:
Summary:
Use us-west-1 resources as part of Devops efforts.

The resources will mirror original state and once things are stable we can turn off original region/resources.

Differential Revision: D49363140


d4l3k (approved):
Lgtm!

PR #773, opened 2023-10-10
Add `aws_batch` dependency extra

schmidt-ai:
Creates a new extra group called `aws_batch` which contains `boto3`. Using the AWS Batch scheduler was failing due to missing `boto3` dependency.

Test plan:

Not sure, open to ideas.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,4 @@
-aiobotocore==2.4.2
+aiobotocore
 ax-platform[mysql]==0.2.3
 black==23.3.0
 boto3


setup.py
```diff
@@ -81,6 +81,7 @@ def get_nightly_version():
             ],
         },
         extras_require={
+            "aws_batch": ["boto3"],
```
kiukchung, line 84: We should probably keep this the same ver as https://github.com/pytorch/torchx/blob/main/dev-requirements.txt#L4 (so `boto3==1.24.59`) or remove the pin in `dev-requirements.txt`

kiukchung (commented):
thanks for the PR!

kiukchung (approved):

kiukchung:
> @kiukchung how would you feel about using a tool such as `pip-tools` or `poetry` to manage dependencies? Manually pinning dev dependencies can be a bit brittle, as this PR demonstrates. I'd be happy to open an issue for this or contribute a PR...

I'd be fine with adding support for poetry as long as we can make the most minimal dev install work (e.g. assuming you have a complian version of python already 1) clone the repo 2) `pip install -e ".[dev]"`)

kiukchung:
@schmidt-ai the CLA Check is failing in CI (there's a link above to sign the CLA).

PR #777, opened 2023-10-17
fixes #775

### LINKED ISSUE #775: metadata support for aws_batch_scheduler
## Description
`AppDef.metadata` is currently not implemented for `aws_batch_scheduler`. This feature request suggests adding it via AWS Batch Tags.

## Motivation/Background
This allows adding metadata that is not available to the training jobs itself (unlike `env` variables). It could be used to filter/group AWs Batch Jobs but won't be pollute `env` variables for the training job.

## Detailed Proposal
Currently there's no way to pass/change AWS Batch job tags for `aws_batch_scheduler` (unlike `env` variables). We can add `AppDef.metadata` key-value pairs as tags in [here](https://github.com/pytorch/torchx/blob/13a4787e0b107cb96d97adf8f9f04e921b58257a/torchx/schedulers/aws_batch_scheduler.py#L515).
This way a custom component can pass the required key-value pairs via `AppDef.metadata` and they will be set on the AWS Batch job as tags.

## Alternatives
- Using `env` variables could result in name clash and, more importantly, leak metadata into the training process itself (e.g. the flag can be used by a dependency).
- Implementing a custom scheduler on top of the standard `aws_batch_scheduler` just to pass custom tags seems like an overkill

## Additional context/links
* https://github.com/pytorch/torchx/blob/13a4787e0b107cb96d97adf8f9f04e921b58257a/torchx/schedulers/aws_batch_scheduler.py#L515

clumsy:
passing AppDef metadata to job tags for AWS Batch scheduler

Test plan:
* updated the unit tests

### DIFF
--- a/torchx/schedulers/aws_batch_scheduler.py
+++ b/torchx/schedulers/aws_batch_scheduler.py
@@ -516,6 +516,7 @@ class AWSBatchScheduler(DockerWorkspaceMixin, Scheduler[AWSBatchOpts]):
                     TAG_TORCHX_VER: torchx.__version__,
                     TAG_TORCHX_APPNAME: app.name,
                     TAG_TORCHX_USER: cfg.get("user"),
+                    **app.metadata,
                 },
             },
             **({"schedulingPriority": priority} if share_id is not None else {}),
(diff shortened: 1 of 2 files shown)


kiukchung (approved):
lgtm thanks for the PR!

clumsy:
Looks like all checks passed, @kiukchung. Please let me know if there's anything I can do to help merging this. Thanks!

PR #781, opened 2023-10-19
(torchx/scheduler)(aws) add instance type for aws_batch_scheduler multinode jobs

### LINKED ISSUE #780: Set container instanceType for multi-node aws_batch jobs
## Description
All node groups in a multi-node parallel jobs must be of the same instance-type, see: https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html#Batch-Type-ContainerProperties-instanceType

## Motivation/Background
AWS Batch Compute Environments can use multiple instance types, see: https://docs.aws.amazon.com/batch/latest/userguide/compute_environment_parameters.html#compute_environment_resources_instanceTypes
By setting `instanceType` containers will only be created on the same requested instance type.

## Detailed Proposal
aws_batch_scheduler.py
```
    if role.num_replicas > 1:
        container["instanceType"] = role.resource.capabilities[K8S_ITYPE]
```

## Alternatives
* Rely on homogeneous AWS Batch Compute Environments

## Additional context/links
* https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html#Batch-Type-ContainerProperties-instanceType
* https://docs.aws.amazon.com/batch/latest/userguide/compute_environment_parameters.html#compute_environment_resources_instanceTypes

clumsy:
add instance type for aws_batch_scheduler multinode jobs

Test plan:
* added two unit tests: one for singlenode and one for multinode job

### DIFF
--- a/torchx/schedulers/aws_batch_scheduler.py
+++ b/torchx/schedulers/aws_batch_scheduler.py
@@ -81,6 +81,7 @@ from torchx.specs.api import (
     runopts,
     VolumeMount,
 )
+from torchx.specs.named_resources_aws import instance_type_from_resource
 from torchx.util.types import none_throws
 from torchx.workspace.docker_workspace import DockerWorkspaceMixin
 from typing_extensions import TypedDict
@@ -244,6 +245,10 @@ def _role_to_node_properties(
         "mountPoints": mount_points,
         "volumes": volumes,
     }
+    if role.num_replicas > 1:
+        instance_type = instance_type_from_resource(role.resource)
+        if instance_type is not None:
+            container["instanceType"] = instance_type
 
     return {
         "targetNodes": f"{start_idx}:{start_idx + role.num_replicas - 1}",
--- a/torchx/specs/named_resources_aws.py
+++ b/torchx/specs/named_resources_aws.py
@@ -29,6 +29,7 @@ Usage:
 
 """
 
+import warnings
 from typing import Callable, Mapping
 
 from torchx.specs.api import Resource
@@ -41,8 +42,20 @@ EFA_DEVICE = "vpc.amazonaws.com/efa"
 # 97% is based on empirical observation that works well for most instance types
 # see: https://docs.aws.amazon.com/batch/latest/userguide/memory-management.html
 MEM_TAX = 0.97
+
+# determines instance type for non-honogeneous CEs
+# see https://github.com/pytorch/torchx/issues/780
 K8S_ITYPE = "node.kubernetes.io/instance-type"
 GiB: int = int(1024 * MEM_TAX)
+
+
+def instance_type_from_resource(resource: Resource) -> str:
+    instance_type = resource.capabilities.get(K8S_ITYPE)
+    if instance_type is None:
+        warnings.warn(
+            "Cannot determine resource instance type which can cause issues for non-homogeneous CEs and multinode jobs. Consider providing torchx.specs.named_resources_aws:K8S_TYPE resource capability."
+        )
+    return instance_type
 
 
 def aws_p3_2xlarge() -> Resource:
--- a/torchx/schedulers/test/aws_batch_scheduler_test.py
+++ b/torchx/schedulers/test/aws_batch_scheduler_test.py
@@ -28,7 +28,9 @@ from torchx.schedulers.aws_batch_scheduler import (
 from torchx.specs import AppState, Resource
 
 
-def _test_app() -> specs.AppDef:
+def _test_app(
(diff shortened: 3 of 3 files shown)


torchx/schedulers/aws_batch_scheduler.py
```diff
@@ -244,6 +245,10 @@ def _role_to_node_properties(
         "mountPoints": mount_points,
         "volumes": volumes,
     }
+    if role.num_replicas > 1:
+        instance_type = role.resource.capabilities.get(K8S_ITYPE, None)
```
kiukchung, line 249: shouldn't the instance type get assigned on the server-side from the resource requirements and the type of instances in the AWS Batch CE?

clumsy: These are the requirements that restrict AWS Batch to only use the specified instance type (CE can have many instance types). For multinode though it's a must for instance type to be the same for all nodes (see the link I attached in the feature request)

clumsy: E.g. if you don't specify it and your CE only operates a single instance type - it will work, but only by chance.

kiukchung: I see, is this specifically to launch between `p4d` and `p4de`? AFAIK there aren't any other accelerated instances that have the same resource specs (cpu, gpu, memMB), so the resource specs should be enough to disambiguate between instance types no?

Couple of comments:

1. We should document in `torchx.specs.named_resources_aws:K8S_TYPE` that this is also used for setting AWS Batch container instance type for non-homogeneous CEs and link issue #780 for details.
2. For custom named resources (see [documentation](https://pytorch.org/torchx/latest/advanced.html#registering-named-resources)) this won't work unless the user remembers to add `K8S_ITYPE` to `Resource.capabilities`  of the custom named resource they are adding. We should update the docs or better yet display a warning (`warnings.warn`) when the (CE is non-homogeneous &&  job is MNP  && resource.capabilities is missing K8S_TYPE) that one should be added to the custom named resource.

clumsy: Right, @kiukchung `p4d` and `p4de` are the two examples of _public_ resource types (e.g. training job requiring precisely `p4de` to get 80GB of GPU memory and if it gets a `p4d` it will OOM).

But it's also applicable to custom resources.

I was looking for a better mechanism to extract a scheduler-specific "metadata". Creating a type hierarchy is perhaps is an overkill, e.g. :
```
class AwsResource(Resource)
    instance_type: str
```

clumsy:
@kiukchung please find an updated commit based on our discussion.

clumsy:
btw lint failure seems unrelated to my change, @kiukchung

torchx/specs/named_resources_aws.py
```diff
@@ -45,161 +46,134 @@
 GiB: int = int(1024 * MEM_TAX)
 
 
+@dataclass
+class AwsResource(Resource):
```
kiukchung, line 50: What you had originally was fine as long as you document that `K8S_ITYPE` is used in the aws_batch_scheduler to set the instance type for non-homogeneous CEs that have instances with resource specs that only differ for accelerators.

kiukchung:
@clumsy let me know if you're blocked on anything here.

clumsy:
@kiukchung just curious if you agree it's better to introduce an AWS resource subclass to provide (self-documented) `instance_type` attribute instead of relying on `K8S_TYPE` behavior that needs to be documented separately. Asking because I'd like to use `instance_type` for my upcoming SageMaker scheduler contribution.
Basically If you're OK with the current PR - I'll address the failures and submit a new version.

kiukchung:
> @kiukchung just curious if you agree it's better to introduce an AWS resource subclass to provide (self-documented) `instance_type` attribute instead of relying on `K8S_TYPE` behavior that needs to be documented separately. Asking because I'd like to use `instance_type` for my upcoming SageMaker scheduler contribution. Basically If you're OK with the current PR - I'll address the failures and submit a new version.

I'd like to avoid introducing subclasses for `Resource` to keep the specs API easier to serialize (e.g. to JSON, YAML, Proto). We can use `K8S_TYPE` (and document that this will be used by the batch and SM schedulers to set instance type params in the respective scheduler's request).

clumsy:
@kiukchung I've made the changes as discussed, the docs build failure does not seem to be caused by my changes.

kiukchung:
LGTM thanks

kiukchung (approved):

PR #783, opened 2023-10-25
Upgrade fsspec and s3fs versions

alanhdu:
Summary: Need to update code for https://github.com/fsspec/filesystem_spec/issues/1394.

Reviewed By: stepanhruda

Differential Revision: D50388544

PR #784, opened 2023-10-30
Update torchx for fsspec 2023.10

alanhdu:
This updates the code to the latest version of fsspec, updating the test to work around https://github.com/fsspec/filesystem_spec/issues/1394.

I have written the test so that it should actually be compatible with *both* the old and new versions of fsspec, but for testing purposes updated the requirements.txt as well.

Supersedes https://github.com/pytorch/torchx/pull/783 (so it can be merged separately from the internal diff).

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@ boto3==1.24.59
 captum>=0.4.0
 docker
 flake8==3.9.0
-fsspec[s3]==2023.1.0
+fsspec[s3]==2023.10.0
 google-api-core
 google-cloud-batch>=0.5.0
 google-cloud-logging>=3.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,7 @@ importlib-metadata
 pyyaml
 docker
 filelock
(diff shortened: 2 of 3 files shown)


kiukchung:
LGTM. Any idea why the unittest: `torchx/schedulers/test/local_scheduler_test.py::LocalDirectorySchedulerTest::test_no_orphan_process_function` is failing?

alanhdu:
I have no idea -- when I tried to run it locally, everything seemed to pass. It looks like when someone also retried the github job and it's passing now?

kiukchung:
> I have no idea -- when I tried to run it locally, everything seemed to pass. It looks like when someone also retried the github job and it's passing now?

@kurman (can you help taking a look at the error logs for kfp?) I retried `KFP Integration Tests` to see if it was something flaky causing the failure, but the retry failed as well.

kiukchung:
merged since KFP Integ tests seem broken in other PRs as well (https://github.com/pytorch/torchx/pull/781, https://github.com/pytorch/torchx/pull/784). Since we just bumped to 0.6.0dev, we can forward fix the test if these PRs continue breaking KFP tests in main.

PR #787, opened 2023-11-02
Add mount args to some `utils` components

### LINKED ISSUE #786: Add `mounts` argument to more `utils` components
## Description
The `python`, `copy`, and maybe even the `touch` util components could benefit from having a `mounts` argument like the `sh` util does.

## Motivation/Background
It's a slight convenience improvement to these utils because they might be used for dealing with e.g. mounted EFS volumes.

## Detailed Proposal
Expand the signature of these util method definitions to include a `mounts` argument, and pass it into the corresponding `Role` constructor.

## Alternatives
Wrap stuff in a call to `utils.sh` with your mounts.

## Additional context/links
Happy to contribute!

schmidt-ai:
Closes #786.

Test plan:
TBD — do we have a good path for testing mounts?


kiukchung:
> Test plan: TBD — do we have a good path for testing mounts?

For components we typically test that the resulting `AppDef` is what we expect it to be. As an example for `mounts` take a look at https://github.com/pytorch/torchx/blob/main/torchx/components/test/dist_test.py#L17. 

Functionally since we reduce components to an `AppDef`, the tests for each scheduler implementation tests for whether `mounts` are respected correctly when we convert the `AppDef` into the respective scheduler job definitions.

PR #791, opened 2023-11-14
updating EFA_DEVICE details for AWS resources

clumsy:
The list of EFA-enabled devices for existing AWS resources is not up-to-date. Adding missing details according to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html#efa-instance-types

Note that the number of devices matches the number of NICs: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#network-cards

Test plan:
N/A

### DIFF
--- a/torchx/specs/named_resources_aws.py
+++ b/torchx/specs/named_resources_aws.py
@@ -137,25 +137,41 @@ def aws_g4dn_4xlarge() -> Resource:
 
 def aws_g4dn_8xlarge() -> Resource:
     return Resource(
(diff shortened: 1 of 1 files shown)


clumsy:

kiukchung:
LGTM, looks like there is a rebase conflict. Happy to merge once you've resolved it!

clumsy:
Bumping this one in case it got slipped through the cracks, @kiukchung

kiukchung:
LGTM thanks for the rebase.

PR #792, opened 2023-11-15
fixing k8s node types for trn1 instance types

clumsy:
Now that we use `K8S_ITYPE` for multi-node job's instance_type for AWS resources, it has to match the official label or we get the following error with `aws_batch_scheduler`:

```
botocore.errorfactory.ClientException: An error occurred (ClientException) when calling the RegisterJobDefinition operation: Error executing request, Exception : Instance type can only be one of [m6g.xlarge, ..., c7g.large, trn1.2xlarge, c7g.12xlarge, ..., trn1.32xlarge, ..., g4dn.4xlarge, trn1n.32xlarge, ..., r7g.medium].

```

Intentionally not keeping an alias to previous labels, e.g. `aws_trn1.32xl` because torchx suggests the new one.

Test plan:
Fixed unit tests

### DIFF
--- a/torchx/components/structured_arg.py
+++ b/torchx/components/structured_arg.py
@@ -191,12 +191,12 @@ class StructuredJArgument:
 
             .. doctest::
 
-                >>> str(StructuredJArgument.parse_from(h="aws_trn1.32xl", j="2"))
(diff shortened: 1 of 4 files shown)


clumsy:
This is a regression, I'm afraid @kiukchung

kiukchung:
LGTM thanks

PR #793, opened 2023-11-16
Fix: Distributed Training Rendezvous error with MCAD v.1.34.1

Sara-KS:
Some releases of MCAD (starting with v1.34.1) incorrectly autogenerated pod labels with the old appwrapper identifier labels added. Since the Kubernetes service deployed by TorchX-MCAD relied on the correct label, distributed jobs failed due to rendezvous errors. This PR moves the generated Kubernetes Service selector to the standard `app.kubernetes.io/instance` label to cover any versions of MCAD that may be in deployment.

Test plan:
Updated the corresponding TorchX-MCAD tests.


kiukchung:
LGTM thanks

PR #794, opened 2023-11-17
Follow python logging best practice in torchx/runner/events/...

gunchu:
Summary: Do not set the log level or filters of either root or module-level loggers. Instead, configure this on each handler.

Reviewed By: AndreasBackx

Differential Revision: D51381654


kiukchung:
LGTM thanks

Issue #802, opened 2023-12-18
Why can't tracker entrypoint be specified in .torchxconfig


clumsy:
### Please note that this issue tracker is not a help form and this issue will be closed.

Before submitting, please ensure you have gone through our
[documentation](https://pytorch.org/torchx).

### Question
The [documentation](https://pytorch.org/torchx/main/tracker.html#user-job-configuration-advanced) is somewhat confusing and is marked for Advanced use after mentioning the mechanism to reference entrypoint, but is there a reason we can't also specify the tracker's entrypoint right in `.torchxconfig` in addition to those discoverable via `entry_points.txt`? E.g.:

```
[torchx:tracker]
my_tracker=my_module:my_function

[tracker:my_tracker]
...
```

clumsy:
If there's no strong reason not to - I can contribute, @kiukchung

kurman:
@clumsy I don't think there is a strong reason, feel free to contribute!

kiukchung:
Answered, closing: entry-point registration is deprecated, so the answer is no longer to put it in `.torchxconfig`. Register a tracker with `@register.tracker()` in a `torchx_plugins.*` namespace package and torchx discovers it; `.torchxconfig` keeps the tracker's configuration, not its entry point. See `docs/source/plugins.rst`.

PR #803, opened 2023-12-19
adding module lookup for building trackers

clumsy:
Discussed this in https://github.com/pytorch/torchx/issues/802

TL;DR this adds the ability to reference a module path instead of entry_point for tracker backend as a fallback when entry_point is not found, e.g.:

```
[torchx.tracker]
my_tracker = torchx.tracker.mlflow.create_tracker

[tracker:my_tracker]
...
```

Test plan:
* Added a unit test, the original functionality is there but as a fallback we also check if we can find tracker backend using module lookup.

### DIFF
--- a/torchx/tracker/api.py
+++ b/torchx/tracker/api.py
@@ -14,6 +14,7 @@ from functools import lru_cache
 from typing import Iterable, Mapping, Optional
 
 from torchx.util.entrypoints import load_group
+from torchx.util.modules import load_module
 
 logger: logging.Logger = logging.getLogger(__name__)
 
@@ -177,30 +178,26 @@ def _extract_tracker_name_and_config_from_environ() -> Mapping[str, Optional[str
 
 
 def build_trackers(
-    entrypoint_and_config: Mapping[str, Optional[str]]
+    factory_and_config: Mapping[str, Optional[str]]
 ) -> Iterable[TrackerBase]:
     trackers = []
 
-    entrypoint_factories = load_group("torchx.tracker")
+    entrypoint_factories = load_group("torchx.tracker") or {}
     if not entrypoint_factories:
-        logger.warning(
-            "No 'torchx.tracker' entry_points are defined. Tracking will not capture any data."
-        )
-        return trackers
+        logger.warning("No 'torchx.tracker' entry_points are defined.")
 
-    for entrypoint_key, config in entrypoint_and_config.items():
-        if entrypoint_key not in entrypoint_factories:
+    for factory_name, config in factory_and_config.items():
+        factory = entrypoint_factories.get(factory_name) or load_module(factory_name)
+        if not factory:
             logger.warning(
-                f"Could not find `{entrypoint_key}` tracker entrypoint. Skipping..."
+                f"No tracker factory `{factory_name}` found in entry_points or modules. See https://pytorch.org/torchx/main/tracker.html#module-torchx.tracker"
             )
             continue
-        factory = entrypoint_factories[entrypoint_key]
         if config:
-            logger.info(f"Tracker config found for `{entrypoint_key}` as `{config}`")
-            tracker = factory(config)
+            logger.info(f"Tracker config found for `{factory_name}` as `{config}`")
         else:
-            logger.info(f"No tracker config specified for `{entrypoint_key}`")
-            tracker = factory(None)
+            logger.info(f"No tracker config specified for `{factory_name}`")
+        tracker = factory(config)
         trackers.append(tracker)
     return trackers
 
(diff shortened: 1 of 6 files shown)


clumsy:
For your consideration, @kurman and @kiukchung.

torchx/tracker/api.py
```diff
@@ -181,26 +183,28 @@ def build_trackers(
...
-                f"Could not find `{entrypoint_key}` tracker entrypoint. Skipping..."
+        factory = entrypoint_factories.get(entrypoint_key)
+        if factory is None:
+            logger.info(
+                f"Could not find `{entrypoint_key}` among entry_points, checking modules..."
             )
-            continue
-        factory = entrypoint_factories[entrypoint_key]
+            try:
+                factory = hydra.utils.get_method(entrypoint_key)
```
kiukchung, line 197: could we not depend on `hydra`?

clumsy: Sure, I didn't notice it's in dev-requirements only. So I take no objections about adding this functionality overall, right @kiukchung ?

clumsy: If not and we don't want to stick to Hydra, we have options for the syntax, e.g. 1) `my_tracker = torchx.tracker.mlflow.create_tracker` (Hydra-ish), 2) `my_tracker = torchx.tracker.mlflow:create_tracker` (entrypoint-ish).
Any preference, @kiukchung?

clumsy:
@kiukchung removed hydra usage, following entrypoint syntax. Please let me know if you have further comments.

torchx/tracker/api.py
```diff
@@ -181,26 +181,27 @@ def build_trackers(
...
+                f"Could not find `{entrypoint_key}` among entry_points, checking modules..."
             )
-            continue
-        factory = entrypoint_factories[entrypoint_key]
+            factory = load_module(entrypoint_key)
+            if not factory:
+                logger.warning(
+                    f"Could not find `{entrypoint_key}` tracker entrypoint. Skipping..."
+                )
+                continue
```
kiukchung, line 199: The logging here should be cleaned up. Rather than logging that you "couldn't" find something, we should run through the fallback logic (first check `entry_points` then `modules`) and at the end, log the result. Something to the effect of:

```
_src = {entry_point or module}

if factory:
   logger.info("Tracker factory `{factory_name}` found in `{_src}`. Creating tracker with config={config}`)
else:
   logger.warning("No tracker factory `{factory_name}` found in entry_point or modules. See: https://pytorch.org/torchx/main/tracker.html#module-torchx.tracker")

```

torchx/tracker/api.py
```diff
@@ -181,26 +181,27 @@ def build_trackers(
...
         if config:
             logger.info(f"Tracker config found for `{entrypoint_key}` as `{config}`")
-            tracker = factory(config)
         else:
             logger.info(f"No tracker config specified for `{entrypoint_key}`")
-            tracker = factory(None)
```
kiukchung, line 203: was this intentional? this is technically BC breaking since `config = {}` (empty map) also evaluates to `False`. In the previous logic, this would end up calling the factory with `None` and in the new logic it would call it with `{}`.

Not a big deal since an empty-map won't cause null check errors, but FYI. Would've been a different story if `factory({})` was changed to `factory(None)` since that would've caused factories without a null-check (e.g. `if config:` or `if config is None:`) to throw.

kiukchung (commented):
Thanks for the PR (and removing the hydra dep). Could you update the docs as well?
Its all done via docstrings in [`torchx/tracker/__init__.py`](https://github.com/pytorch/torchx/blob/main/torchx/tracker/__init__.py#L36)

You'll want to update the `Tracker Setup` and split out the rest of the docs into `entry_point` and `module` registrations.

You can preview the docs locally by:

```
$ cd docs
$ pip install -r requirements.txt
$ make clean html
$ python -m http.server -d build/0.7.0.dev0/html
```
Then opening up the docs page on `https://localhost:8080`

see [docs-build](https://github.com/pytorch/torchx/blob/main/.github/workflows/doc-build.yaml) CI action for details.

torchx/tracker/api.py
clumsy: Not intentional, I don't even have this locally, weird. I'll make sure it's not here and no BC is introduced.

clumsy:
Ok @kiukchung I addressed the comments, updated the documentation. Additionally I extracted the `load_module` functionality and tests to be separate from entrypoints. I didn't test the documentation rendering yet.

clumsy:
I addressed an unrelated check failure due a broken link in Kubernetes docs. KPF seem to fail for an unrelated reason, @kiukchung.

kiukchung (approved):

clumsy:
@kiukchung do I need to make furhter changes to this PR? I don't have conflicts when fetch/rebasing on top of `origin/main` locally.

PR #810, opened 2024-01-24
feat: add job_role_arn to aws_batch_scheduler

clumsy:
In order to let Batch container access AWS resources in a different account we need to supply [jobRoleArn](https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html#Batch-Type-ContainerProperties-jobRoleArn) for the container.

Test plan:
☑ added a unit test
☑ submitted a job to batch


clumsy:

kiukchung (approved):
lgtm thanks!

PR #817, opened 2024-02-07
Extract env set up/output handling in local_scheduler for easier subclassing

cniii:
Summary:
## No functional change
Extracted stdout, stderr, and env var handling logic from `local_scheduler`'s `_popen` for easier subclass function overrides.

Differential Revision: D53453899


manav-a:
Just triggered the tests lets see how things go

manav-a (approved):
Looks good to me. The tests failures here are pre-existing ones and we will look at them independently

PR #818, opened 2024-02-07
upgrade pyre version in `fbcode/torchx` - batch 1

manav-a:
Differential Revision: D50848925


yikaiMeta:
Thanks for fixing this @manav-a

PR #822, opened 2024-02-14
Added Kueue Job Scheduler

Bobbins228:
Added a Kubernetes Kueue batch job scheduler based on the Kubernetes Scheduler
Note: variable `local_kueue="local-kueue-name"` is required in the scheduler args for the `queue-name` label and for priority add `kueue_priority_class="kueue-priority-class-name"` to the sheduler args

## Testing
Set up a Kubernetes/Openshift cluster
* Install [Kueue](https://kueue.sigs.k8s.io/docs/installation/#install-a-released-version) 
    * Create necessary Kueue resources: [Cluster Queue](https://kueue.sigs.k8s.io/docs/concepts/cluster_queue/), [Resource Flavour](https://kueue.sigs.k8s.io/docs/concepts/resource_flavor/#empty-resourceflavor), [Local Queue](https://kueue.sigs.k8s.io/docs/concepts/local_queue/) and [workload priority classes](https://kueue.sigs.k8s.io/docs/concepts/workload_priority_class/) for testing job priority
* Clone this branch locally and create a torchx whl file with `python3 setup.py bdist_wheel`
* Install the whl file with `pip install dist/torchx-0.7.0.dev0-py3-none-any.whl`
* Authenticate to your K8s cluster
* Run a sample job with ```torchx run --scheduler kueue_job --scheduler_args namespace=default,local_kueue="default-kueue",image_repo="user/alpine" utils.echo --image alpine:latest --msg hello``` - should return something like `kueue_job://torchx_user/1234`
* Get the jobs status with `torchx status kueue_job://torchx_user/1234`
* All jobs created with this scheduler should gain the `Suspended`/`JobResumed` status 

Test plan:

Created Unit tests based on Kubernetes Unit tests.


torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,809 @@
...
+        requests["cpu"] = f"{request_mcpu}m"
+    if resource.memMB > 0:
+        limits["memory"] = f"{int(resource.memMB)}M"
+        request_memMB = max(int(resource.memMB) - RESERVED_MEMMB, 0)
+        requests["memory"] = f"{request_memMB}M"
+    if resource.gpu > 0:
+        requests["nvidia.com/gpu"] = limits["nvidia.com/gpu"] = str(resource.gpu)
+
+    for device_name, device_limit in resource.devices.items():
+        limits[device_name] = str(device_limit)
```
KPostOffice, line 187: Why aren't limits set here as they are with CPU, Mem, and GPU?

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,809 @@
...
+    # 64M default so we mount an unlimited sized tmpfs directory on it.
+    SHM_VOL = "dshm"
+    volumes = [
+        V1Volume(
+            name=SHM_VOL,
+            empty_dir=V1EmptyDirVolumeSource(
+                medium="Memory",
+            ),
+        ),
+    ]
```
KPostOffice, line 208: Should we be worried about limiting the size of the shared memory dir? By default it can use up to 50% of the total allocated memory for the pod.

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,809 @@
...
+        # map any local images to the remote image
+        images_to_push = self.dryrun_push_images(app, cast(Mapping[str, CfgVal], cfg))
+
+        service_account = cfg.get("service_account")
+        assert service_account is None or isinstance(
+            service_account, str
+        ), "service_account must be a str"
+
+        local_kueue = cfg.get("local_kueue")
+        assert local_kueue is None or isinstance(
```
KPostOffice, line 564: You can remove `local_kueue is None` from this conditional that way you don't need to check for it in `app_to_resource`

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,809 @@
...
+    def _cancel_existing(self, app_id: str) -> None:
+        from kubernetes import client
+        namespace, name = app_id.split(":")
+
+        self._batchv1_api().delete_namespaced_job(
+            namespace=namespace,
+            name=name,
+            body=client.V1DeleteOptions(
+                    propagation_policy='Foreground',
+                    grace_period_seconds=5))
```
KPostOffice, line 593: What's the reason behind not just using the default `grace_period`?

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,809 @@
...
+            return f"Exception: {e}"
+        
+
+        try: 
+            status = job.status
+        except:
+            status = None
+        
+        if status:
+            condition = status.conditions[0]
```
KPostOffice, line 640: Is the first element of conditions special somehow?

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,809 @@
...
+            if condition.type == "Suspended" and condition.reason == "JobSuspended":
+                app_state = JOB_STATE["Suspended"]
+            elif condition.reason == "JobResumed":
+                if status.active is not None:
+                    app_state = JOB_STATE["Running"]
+                elif status.succeeded is not None:
+                    app_state = JOB_STATE["Completed"]
+                elif status.failed is not None:
+                    app_state = JOB_STATE["Failed"]
```
KPostOffice, line 649: From my understanding of https://docs.openshift.com/container-platform/4.14/rest_api/workloads_apis/job-batch-v1.html#status, `Completed`, `Running`, and `Failed` will be integer values representing the number of pods in each state. As it stands, if there are 2 pods that succeed and 2 that fail your app_state would be `Completed`.

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,809 @@
...
+                else:
+                    app_state = JOB_STATE["Pending"]
+            else:
+                app_state = JOB_STATE["Pending"]
+
+            for condition in status.conditions or []:
+                role, _, idx = condition.type.rpartition("-")
```
KPostOffice, line 656: According to https://docs.openshift.com/container-platform/4.14/rest_api/workloads_apis/job-batch-v1.html#status-conditions-2 `condition.type` is either `Succeeded` or `Failed` so this wouldn't work for getting the role.

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,809 @@
...
+                "KueueJobScheduler only supports COMBINED log stream"
+            )
+
+        from kubernetes import client, watch
+
+        namespace, name = app_id.split(":")
+
+        pod_name = get_pod_name_from_job(self, job_name=name, namespace=namespace)
+        if pod_name is None:
+            return "Pods not found. Is the Job Suspended?"
```
KPostOffice, line 707: I don't think returning a string is appropriate here. I would either raise an error or log an exception and return an empty list.

torchx/schedulers/kueue_job_scheduler.py
Bobbins228: I see so a job could be considered a failure and a success with the current set up. I will look into this and come up with another way of gauging the status :+1:

torchx/schedulers/kueue_job_scheduler.py
Bobbins228: I was testing something and forgot to remove the grace periods that can go :+1:

torchx/schedulers/kueue_job_scheduler.py
Bobbins228: It contains the `Suspended` condition that is applied by Kueue.
```
'conditions': [{'last_probe_time': datetime.datetime(2024, 2, 15, 12, 3, 38, tzinfo=tzutc()),
                 'last_transition_time': datetime.datetime(2024, 2, 15, 12, 3, 38, tzinfo=tzutc()),
                 'message': 'Job resumed',
                 'reason': 'JobResumed',
                 'status': 'False',
                 'type': 'Suspended'},
                {'last_probe_time': datetime.datetime(2024, 2, 15, 12, 6, 23, tzinfo=tzutc()),
                 'last_transition_time': datetime.datetime(2024, 2, 15, 12, 6, 23, tzinfo=tzutc()),
                 'message': None,
                 'reason': None,
                 'status': 'True',
                 'type': 'Complete'}],
```

ccharest93:
What does kueue do that volcano doesn't? Is there a reason to support more than one batch scheduler on kubernetes?

Bobbins228:
Hi @ccharest93,
We have decided to add a Kueue Batch Job scheduler as Kueue is backed by Kubernetes Batch Sig and this gives users more options in terms of choosing a Kubernetes Scheduler.

scripts/component_integration_tests.py
```diff
@@ -94,6 +95,14 @@ def main() -> None:
                 "namespace": "torchx-dev",
             },
         },
+        "kueue_job": {
+            "providers": [
+                component_provider,
+                examples_app_defs_providers,
+            ],
+            "image": torchx_image,
+            "cfg": {"namespace": "torchx-dev", "local_kueue": "default-kueue"},
```
KPostOffice, line 104: I believe this needs to be `local_queue`

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,820 @@
...
+
+        service_account = cfg.get("service_account")
+        assert service_account is None or isinstance(
+            service_account, str
+        ), "service_account must be a str"
+
+        local_queue = cfg.get("local_queue")
+        assert isinstance(
+            local_queue, str
+        ), "local_queue is a required string please specify local_kueue in scheduler_args"
```
KPostOffice, line 562: this should be `local_queue` as well

KPostOffice (commented):
One quick comment

torchx/schedulers/__init__.py
```diff
@@ -22,6 +22,7 @@
     "gcp_batch": "torchx.schedulers.gcp_batch_scheduler",
     "ray": "torchx.schedulers.ray_scheduler",
     "lsf": "torchx.schedulers.lsf_scheduler",
+    "kueue_job": "torchx.schedulers.kueue_job_scheduler"
```
d4l3k, line 25: can we use just `kueue` and `kueue_scheduler` to keep it inline with other schedulers? Seems like the project name is just Kueue rather than Kueue Job https://github.com/kubernetes-sigs/kueue

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,820 @@
...
+    # releasing and recreating
+    "Restarting": AppState.RUNNING,
+    # Completed is the phase that all tasks of Job are completed successfully
+    "Completed": AppState.SUCCEEDED,
+    # Terminating is the phase that the Job is terminated, waiting for releasing
+    # pods
+    "Terminating": AppState.RUNNING,
+    # Teriminated is the phase that the job is finished unexpected, e.g. events
+    "Terminated": AppState.FAILED,
+    "Failed": ReplicaState.FAILED,
```
d4l3k, line 105: this is mixing AppStates and ReplicaStates -- is that intentional?

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,820 @@
...
+
+
+def sanitize_for_serialization(obj: object) -> object:
+    from kubernetes import client
+
+    api = client.ApiClient()
+    return api.sanitize_for_serialization(obj)
+
+
+def role_to_pod(name: str, role: Role, service_account: Optional[str]) -> "V1Pod":
```
d4l3k, line 144: can we share this logic with the Volcano scheduler?

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,820 @@
...
+            type_=str,
+            help="The service account name to set on the pod specs",
+        )
+        opts.add(
+            "local_queue",
+            type_=str,
+            help="The Local Kueue name to set on the local Kueue label",
+        )
+        opts.add(
+            "kueue_priority_class",
```
d4l3k, line 617: can drop the `kueue_` prefix since all of these are Kueue fields

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,820 @@
...
+            print(f"Cannot gather job status: {e}")
+            status = None
+
+        if status:
+            for condition in status.conditions or []:
+                role, _, idx = job.metadata.name.rpartition("-")
+                if condition.type == "Suspended":
+                    if condition.reason == "JobSuspended":
+                        app_state = JOB_STATE["Suspended"]
+                        state = KUEUE_STATE["JobSuspended"]
```
d4l3k, line 652: can we index into these maps with the reason instead of using lots of if/else?

torchx/specs/api.py
```diff
@@ -404,6 +404,8 @@ class AppState(int, Enum):
     6. FAILED - app has unsuccessfully completed
     7. CANCELLED - app was cancelled before completing
     8. UNKNOWN - app state is unknown
+    9. SUSPENDED - app is suspended
+    10. Resumed - app is resumed
```
d4l3k, line 408: nit: use same capitalization

scripts/component_integration_tests.py
```diff
@@ -94,6 +95,14 @@ def main() -> None:
                 "namespace": "torchx-dev",
             },
         },
+        "kueue_job": {
```
d4l3k, line 98: can we add an integration test that uses minikube similar to the existing one for KFP? 

ex: https://github.com/pytorch/torchx/pull/696

d4l3k (changes requested):
Exciting to see this -- does this also work with the native K8s batch API?

torchx/schedulers/kueue_job_scheduler.py
```diff
@@ -0,0 +1,820 @@
+#!/usr/bin/env python3
```
d4l3k, line 1: do we need to keep this file? seems like kueue_scheduler is the same?

scripts/setup_minikube_kueue.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
```
d4l3k, line 1: can we add a integration test for this similar to the minikube one for volcano?

You should be able to push to a draft PR (or this one) and test it e2e 

Similar setup to https://github.com/pytorch/torchx/blob/main/.github/workflows/components-integration-tests.yaml#L79-L82

torchx/util/role_to_pod.py
```diff
@@ -0,0 +1,183 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+from torchx.specs.api import (
+    BindMount,
+    DeviceMount,
+    Role,
+    VolumeMount,
+)
+from typing import Dict, Optional
+from kubernetes.client.models import V1Pod
```
d4l3k, line 13: can we wrap this in a TYPE_CHECKING block? we late import kubernetes to avoid adding long import times. 

Though we do have other changes that makes this a bit less necessary

torchx/schedulers/kueue_scheduler.py
```diff
@@ -0,0 +1,648 @@
...
+
+        try:
+            api_instance = self._batchv1_api()
+            job = api_instance.read_namespaced_job_status(name, namespace)
+        except client.ApiException as e:
+            return f"Exception: {e}"
+        try:
+            status = job.status
+        except Exception as e:
+            print(f"Cannot gather job status: {e}")
```
d4l3k, line 474: can we use logger? stdout is reserved for certain programmatic outputs

d4l3k:
Looks like there's also valid pyre/lint issues

torchx/schedulers/kueue_job_scheduler.py
Bobbins228: That is odd I don't know how that snuck in there I will remove :+1:

d4l3k:
@Bobbins228 there's been a bunch of cleanup on tests in on the `main` branch -- if you rebase/merge all tests should be passing (assuming no new breaking changes)

xujyan:
@Bobbins228 did you try multi-node ddp with it? I think you'd need a JobSet for that which sets up a headless k8s service for workers to reach the master. See example here https://github.com/kubernetes-sigs/jobset/blob/main/examples/pytorch/resnet-cifar10/resnet.yaml

Another way to approach this is perhaps to not be tightly coupled with a specific operator but rather just output a k8s standard Job (or even Pod) spec with the container image and allow it to be piped into other tools that can transform it into whatever that's suitable. But AFAICT something like a `--output-format yaml` isn't supported in torchx and `--dryrun` aims to provide human readable instead of tool parsable output, so I don't know if this is realistic?

d4l3k:
@xujyan @Bobbins228 looks like the ddp job on Kueue is having issues

xujyan:
I would expect that. Like I suggested above you'd need to produce a `JobSet` instead of a `Job`, and as the example [here](https://github.com/kubernetes-sigs/jobset/blob/main/examples/pytorch/resnet-cifar10/resnet.yaml) shows, point to the master at 

```
              - name: MASTER_ADDR
                value: "pytorch-workers-0-0.pytorch"
```

where the value is of the format `<jobset_name>-<job_name>-0-0.<jobset_name>`

xujyan:
And you'd need to install jobset controller in the test env too, see [README](https://github.com/kubernetes-sigs/jobset)

PR #830, opened 2024-02-24
[torchx][runner] Simplify Runner.close.

ttsugriy:
there is no need to use item tuples since `name` is not used

Test plan:
pyre -l torchx/runner/api.py


manav-a (approved):
Thanks for doing this

PR #832, opened 2024-02-27
Update code format in api_test.py

yikaiMeta:
change code format only

Test plan:


yikaiMeta:
the change in this PR has landed internally. So no need to import.

manav-a (approved):
Looks good to me

PR #835, opened 2024-02-29
fix slurm integration tests

d4l3k:
Integration test was running out of disk space. This uses a larger Linux worker which should have more disk.

Test plan:

CI


manav-a (approved):
Awesome thanks for fixing this!

PR #841, opened 2024-03-04
Update local_scheduler.py

yikaiMeta:
apply patch.
On meta side the auto patch fails. So I have to manually apply patch here.

Test plan:


yikaiMeta:
The next time linter fails, we should be careful as this could mean the format is wrong. And once it's landed, diff train might be broken.

manav-a (approved):
Looks good to me

PR #843, opened 2024-03-04
Print format suggestions from Linter on OSS.

yikaiMeta:
Summary:
The linter should print more details when the format doesn't align with what black+usort require.
This diff prints the format suggestions on OSS side.

Differential Revision: D54502339

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -76,8 +76,23 @@ fi
 CHANGED_FILES="$(git diff --name-only | grep '\.py$' | tr '\n' ' ')"
 if [ "$CHANGED_FILES" != "" ]
 then
+    RED="\e[31m"
+    ENDCOLOR="\e[0m"
+    echo "------------------------------------------"
+    echo -e "${RED}[format] These files are not well-formatted:${ENDCOLOR}"
     git diff --name-only
-    echo "There are uncommitted changes on disk likely caused by the linters."
+    echo "------------------------------------------"
+    echo -e "${RED}[format] Suggested format by lint:${ENDCOLOR}"
+    git diff
+    echo "------------------------------------------"
+    echo "To apply the suggested format, run:"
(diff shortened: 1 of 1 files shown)


scripts/lint.sh
```diff
@@ -76,8 +76,14 @@ fi
...
 then
+    RED="\033[0;31m"
+    echo "-------------------------------"
+    echo "${RED} [format] These files are not well-formatted:"
     git diff --name-only
-    echo "There are uncommitted changes on disk likely caused by the linters."
+    echo "-------------------------------"
+    echo "${RED} [format] Suggested format by lint:"
+    git diff
+    echo "${RED} If these changes will be synced to Meta's codebase, you must fix them because code format is enforced in Meta."
```
d4l3k, line 86: aren't all changes synced? We generally prefer to avoid Meta specific comments/changes so would prefer to just remove this line

d4l3k (approved):
lgtm

scripts/lint.sh
yikaiMeta: that makes sense. Let me remove it.

scripts/lint.sh
```diff
@@ -76,8 +76,14 @@ fi
 CHANGED_FILES="$(git diff --name-only | grep '\.py$' | tr '\n' ' ')"
 if [ "$CHANGED_FILES" != "" ]
 then
+    RED="\033[0;31m"
+    echo "-------------------------------"
+    echo "${RED} [format] These files are not well-formatted:"
     git diff --name-only
-    echo "There are uncommitted changes on disk likely caused by the linters."
+    echo "-------------------------------"
+    echo "${RED} [format] Suggested format by lint:"
```
manav-a, line 84: Can we also suggest the commands that can be run locally for sure use

manav-a (approved):

PR #845, opened 2024-03-06
Pyre Configurationless migration for] [batch:231/244]

connernilsen:
Differential Revision: D54562955

### DIFF
--- a/scripts/collect_env.py
+++ b/scripts/collect_env.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+# pyre-strict
+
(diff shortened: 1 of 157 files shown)


manav-a (approved):
Looks good to me

manav-a:
Going forward do we need to add strict to files? Anything preventing us from just doing this in the pyre config?

connernilsen:
Copying the response I gave internally:

>The short version: Pyre is going through some architectural changes that will require us to stop using .pyre_configuration.local files. The benefits we'll get for doing so are huge, but that means settings such as the strictness of a project need to be migrated into the files themselves if we want to type check them, especially if we want CI, CLI, and IDE type check runs to return the same results.

PR #846, opened 2024-03-07
Accept non-string complex values in defaults argument in materialize_appdef 

ishachirimar:
1. modify torchx/utils/types to have a `decode` function that takes in a value and annotation and based on the annotation, returns a decoded version of the value. if the value is not encoded when passed in the same value is returned. 
2. call this function in `materialize_appdef` instead of previous conditional calling of various decode functions. 
3. add unit tests for `decode` and `materialize_appdef`


yikaiMeta (approved):

yikaiMeta:
this test needs attention: https://github.com/pytorch/torchx/actions/runs/8192055115/job/22402711927
I've rerun it. Let's see if it can pass.

PR #853, opened 2024-03-21
suppress errors in `torchx`

kurman:
Differential Revision: D55109047


manav-a (approved):
Looks good to me

d4l3k (changes requested):
this field has only been removed on latest PyTorch which is causing this issue

kiukchung:
The corresponding internal change (D55109047) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #854, opened 2024-03-22
Make torchx error out when the same arg is passed in twice

andywag:
Summary: Added an action to the arg parser to catch this condition and error out

Differential Revision: D55216738


manav-a:
Running tests on the change

torchx/cli/cmd_run.py
```diff
@@ -113,6 +113,23 @@ def run(self, args: argparse.Namespace) -> None:
             print(get_builtin_source(builtin_name))
 
 
+class SchedulerArgOnce(argparse.Action):
+    scheduler_args_set = False
```
manav-a, line 117: Do we need this?

torchx/cli/cmd_run.py
```diff
@@ -113,6 +113,23 @@ def run(self, args: argparse.Namespace) -> None:
             print(get_builtin_source(builtin_name))
 
 
+class SchedulerArgOnce(argparse.Action):
```
manav-a, line 116: Maybe we can reuse this for other commands / options as well?

torchx/cli/cmd_run.py
andywag: Yes, it's a one-shot which is set to true when this arg is parsed. This can be converted to a set containing the arg so the class can be generalized to work for more cases.

torchx/cli/cmd_run.py
andywag: Yes or at least the logic can be used. The same logic can also be added to the _torchxconfig class which is an argparse.Action used for other inputs. This should cover all the cases except the "store_true" cases where the error is less meaningful.

yikaiMeta (approved):

PR #855, opened 2024-03-23
schedulers/docker: Add support for setting environment variables

asuta274:
This change adds support for ``Dict[str, str]`` type in ``runopts`` and use it to expose env flag in ``DockerScheduler``.
The use-case is for setting environment variables for ``local_docker`` scheduler in order to access some remote resources.

Test plan:

```
pytest torchx/schedulers/test/docker_scheduler_test.py
pytest torchx/specs/test/api_test.py
pytest torchx/runner/test/config_test.py
```

### DIFF
--- a/torchx/runner/test/config_test.py
+++ b/torchx/runner/test/config_test.py
@@ -143,6 +143,8 @@ _TEAM_CONFIG = """#
 s = team_default
 i = 50
 f = 1.2
+d = a:b,c:d
+d_none= x:y
 """
 
 _MY_CONFIG = """#
@@ -368,6 +370,8 @@ image = foobar_custom
             self.assertEqual("runtime_value", cfg.get("s"))
             self.assertEqual(50, cfg.get("i"))
             self.assertEqual(1.2, cfg.get("f"))
+            self.assertEqual({"a": "b", "c": "d"}, cfg.get("d"))
+            self.assertEqual({"x": "y"}, cfg.get("d_none"))
 
     @patch(
         TORCHX_GET_SCHEDULER_FACTORIES,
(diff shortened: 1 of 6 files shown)


torchx/runner/test/config_test.py
```diff
@@ -105,6 +105,18 @@ def _run_opts(self) -> runopts:
...
+            "d",
+            type_=Dict[str, str],
+            default={"foo": "bar"},
+            help="a dict option",
+        )
+        opts.add(
+            "d_none",
+            type_=Dict[str, str],
+            default=None,
+            help="a None dict option",
```
d4l3k, line 118: can we add a test that use these opts below to verify parsing is correct?

torchx/schedulers/docker_scheduler.py
```diff
@@ -359,6 +364,12 @@ def _run_opts(self) -> runopts:
             default=None,
             help="list of glob patterns of environment variables to copy if not set in AppDef. Ex: FOO_*",
         )
+        opts.add(
+            "env",
+            type_=Dict[str, str],
+            default=None,
+            help="environment varibles to be passed to the run (e.g. ENV1:v1,ENV2:v2,ENV3:v3)",
```
d4l3k, line 371: can we document that either , or ; will work?

nit: spelling `variables`

d4l3k: may also be good to document that this will apply on top of the copied envs

d4l3k (approved):
LGTM with some small comments

There's some pyre and lint errors -- do you mind fixing them? Thanks

torchx/runner/test/config_test.py
asuta274: Done

torchx/schedulers/docker_scheduler.py
asuta274: Done

asuta274:
> LGTM with some small comments
> 
> There's some pyre and lint errors -- do you mind fixing them? Thanks

Done!

PR #865, opened 2024-03-28
Support home directory sign (~) in BindMount path

asuta274:
This PR is to support home directory sign (~) in BindMount path. 

Test plan:

Unit tests
```
pytest torchx/specs/test/builders_test.py
```

Manually tested with local_docker and the config like below:
```
[component:dist.ddp]
mounts=type=bind,src=~/.config/gcloud/application_default_credentials.json,dst=/creds.json
```
``torchx run`` would fail before the change and work after.


d4l3k (approved):
LGTM -- thanks for contributing!

PR #867, opened 2024-03-29
fix nightly.yaml

yikaiMeta:
Test plan:

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -13,7 +13,7 @@ jobs:
       - name: Setup Python 3.10
         uses: actions/setup-python@v2
         with:
-          python-version: 3.10
+          python-version: '3.10'
           architecture: x64
       - name: Checkout TorchX
         uses: actions/checkout@v2
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -4,7 +4,6 @@ on:
   # run every day at 11:15am
   schedule:
     - cron:  '15 11 * * *'
-  pull_request:
 
 jobs:
   nightly:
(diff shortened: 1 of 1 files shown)


yikaiMeta:
the default Linux github-hosted runner has 2 GPU, 7GB memory and 14 GB storage.
Here is github's official doc: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners
Haven't found the explanation of linux.20_04.4x but lots of our github Action yaml files are using it.

yikaiMeta:
Push nightly job runs all tests. Two tests fail.

.github/workflows/nightly.yaml
```diff
@@ -22,7 +23,7 @@ jobs:
           pip install -e .[dev]
           pip install twine
       - name: Run tests
-        run: python -m unittest discover --verbose --start-directory . --pattern "*_test.py"
+        run: pytest --verbose -o python_files="*_test.py"
```
d4l3k, line 26: The filter isn't necessary I don't think

Our main python tests just call `pytest`: https://github.com/pytorch/torchx/blob/9f20e5a55233288a3ba6fb9bec4e72d1e5fbf6c2/.github/workflows/python-unittests.yaml#L40

.github/workflows/nightly.yaml
```diff
@@ -4,15 +4,16 @@ on:
   # run every day at 11:15am
   schedule:
     - cron:  '15 11 * * *'
+  pull_request:
```
d4l3k, line 7: please remove before landing

d4l3k (approved):
generally looks good to me -- think the test failures are on main so that's expected

ryxli:
@yikaiMeta 

Looks like nightly push runs still fail?
https://github.com/pytorch/torchx/actions/runs/8553750453/job/23437562048#step:5:13

Latest nightly release is from [2024.2.12](https://pypi.org/project/torchx-nightly/)

 
Believe `pytest-cov` needs to be added to [dev-requirements.txt](https://github.com/pytorch/torchx/blob/main/dev-requirements.txt)

PR #868, opened 2024-03-29
Fixed Multiple Args Testing Issue

andywag:
Summary: The purpose of the test was to check removal of multiple args. It passed internally using a facebook option but failed externally because the option didn't exist.

Differential Revision: D55525717


d4l3k (approved):

d4l3k:
Regression was introduced in https://github.com/pytorch/torchx/pull/854

PR #870, opened 2024-03-30
Make torchx scheduler opts support enums

andywag:
Summary:
Added Support for Enumerations in scheduler options.

This is a bit more generic as it takes in a creator function which converts a CfgVal to another type. There are some limitations on how general it can be based on how the typing is setup. This also makes it tricky to make a convenience function to handle only Enums.

Differential Revision: D55551233

### DIFF
--- a/torchx/specs/api.py
+++ b/torchx/specs/api.py
@@ -702,6 +702,7 @@ class runopt:
     opt_type: Type[CfgVal]
     is_required: bool
     help: str
+    creator: Optional[Callable[[CfgVal], CfgVal]] = None
 
 
 class runopts:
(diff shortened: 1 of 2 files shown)


yikaiMeta (approved):

torchx/specs/api.py
```diff
@@ -702,6 +702,7 @@ class runopt:
     opt_type: Type[CfgVal]
     is_required: bool
     help: str
+    creator: Optional[Callable[[CfgVal], CfgVal]] = None
```
kiukchung, line 705: Its not clear how to use this `creator`. Can/should I use a `creator` for a primitive `CfgVal` (eg ones that are not enums)?

In general schedule options were kept dead simple because they are meant to be directly parsable from command-line (after all torchx is mostly used from CLI). This is why they are not structured, and only basic type checking is done for sanity. Enums are tricky to support especially when the values are sent over the wire and as such support for enums in scheduler opts should be well motivated and thorougly justified as we would be trading off simplicity/clarity for <whatever the motivation is>.

I'd like to understand the motivation behind wanting to support enums for scheduler opts before we add them. cc @d4l3k

kiukchung (changes requested):

kiukchung:
The corresponding internal change (D55551233) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #873, opened 2024-04-08
Adding health check settings to Role spec

gag1jain:
Summary:
Spec for health check settings for a Role, which can be used for enabling health check of the node.

Health check settings:
```
port: port number for health check
regex: regex for health check success
timeout_secs: timeout for health check
interval_secs: interval between health checks
fail_secs: time to fail after health check failure
startup_grace_secs: startup grace period
```

Health check settings can be used by different schedulers to enable corresponding health check mechanism. By default health check is disabled.

Differential Revision: D55876399

### DIFF
--- a/torchx/specs/api.py
+++ b/torchx/specs/api.py
@@ -280,6 +280,27 @@ class DeviceMount:
 
 
 @dataclass
+class HealthCheck:
+    """
+    Defines health check settings for a node
+    Args:
+        port: port number for health check
+        regex: regex for health check success
+        timeout_secs: timeout for health check
+        interval_secs: interval between health checks
+        fail_secs: time to fail after health check failure
+        startup_grace_secs: startup grace period
+    """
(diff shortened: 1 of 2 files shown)


kiukchung:
will health checks be implemented for the majority of existing scheduler implementations? If not (and this is specific to a particular scheduler - eg mast) could we think of a way to extend the particular scheduler with health check configs (eg through structured metadata) rather than adding this to the general spec api. Otherwise the spec api will start accumulating a lot of seemingly unimplemented specs.

d4l3k:
@gag1jain this may be better to add to https://www.internalfb.com/code/fbsource/[edit]/fbcode/torchx/specs/fb/api_extended.py rather than the public Role definition

d4l3k (changes requested):
would prefer to have this be in api_extended.py if possible

PR #880, opened 2024-04-11
Remove TORCHX_COMPONENT_ARGS env var from role

ishachirimar:
Summary: remove torchx component args env var to avoid taking too much storage/ hitting limits for env var size

Differential Revision: D56033363

### DIFF
--- a/torchx/runner/api.py
+++ b/torchx/runner/api.py
@@ -324,6 +324,10 @@ class Runner:
 
         configured_trackers = get_configured_trackers()
 
+        app_handle = make_app_handle(scheduler, self._name, macros.app_id)
+
+        os.environ[ENV_TORCHX_JOB_ID] = app_handle
+
         for role in app.roles:
             if not role.entrypoint:
                 raise ValueError(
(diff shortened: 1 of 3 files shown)


torchx/runner/api.py
```diff
@@ -324,6 +324,10 @@ def dryrun(
 
         configured_trackers = get_configured_trackers()
 
+        app_handle = make_app_handle(scheduler, self._name, macros.app_id)
+
+        os.environ[ENV_TORCHX_JOB_ID] = app_handle
```
kiukchung, line 329: what is the purpose of setting the `app_handle` in the current process' environment variables?

kiukchung (changes requested):
Can you elaborate on what storage limit you are hitting with env vars? Simply removing 1 env var solves the problem? Looks like the intention here is really to revert https://github.com/pytorch/torchx/commit/eb0781d438bc5b649054a3fdd14d77f77e8ed6ce. If so can you submit a PR as a `git revert`?

I'm assuming you're hitting a limit in the scheduler (mast) frontend that limits the number of env vars the role spec can have.

There's another bundled change here that modifies the TORCHX_JOB_ID env var key in the tracker - that should be a separate PR for better visibility and history tracking.

torchx/runner/api.py
ishachirimar: Hi there, i think i exported the wrong version of my changes to a PR. Let me revise this and update.

PR #888, opened 2024-04-18
Fix for issue with multiple -- in command line

andywag:
Summary: Repeated args in the command line were removed to avoid errors but a condition with repeated -- caused a breakage

Differential Revision: D56317273


hstonec (approved):
LGTM

PR #890, opened 2024-04-22
add workspace to TorchXEvent for logging

ishachirimar:
Summary:
add the workspace to the TorchXEvent type, so eventually it can be stored to the tsm logger.

this will let us track how different workspaces/ no workspace affects build and scheduler

Differential Revision: D56324618


kiukchung:
The corresponding internal change (D56324618) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #891, opened 2024-04-23
add workspace to TorchXEvent for logging

ishachirimar:
Summary:
add the workspace to the TorchXEvent type, so eventually it can be stored to the tsm logger.

this will let us track how different workspaces/ no workspace affects build and scheduler

Differential Revision: D56324592


manav-a (approved):
Looks good to me

PR #892, opened 2024-04-23
Migrate components-integration-tests to Terraform

yikaiMeta:
In this PR, all the AWS resources that components-integration-tests are using will be from Terraform.
The IAM role is migrated to TF version.
The ECR is migrated to the container registry in US-west-1, in which all resources are from Terraform.
The kubernetes scheduler in this test is using a minicube, which doesn't require AWS EKS. So I deleted the EKS setup.

Test plan:
Components Integration Tests can pass.

### DIFF
--- a/.github/workflows/components-integration-tests.yaml
+++ b/.github/workflows/components-integration-tests.yaml
@@ -11,10 +11,10 @@ jobs:
     strategy:
       matrix:
         include:
-          - scheduler: "aws_batch"
-            platform: "linux.20_04.4x"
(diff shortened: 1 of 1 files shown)


andywag (approved):

d4l3k (approved):
LGTM -- looks like it ran with secrets

manav-a:
Just for my understanding just changing the terraform here runs the CI with the change?

yikaiMeta:
> Just for my understanding just changing the terraform here runs the CI with the change?

yes that's correct.

Issue #897, opened 2024-04-26
privileged flag for local_docker scheduler


clumsy:
## Description
It is available for `aws_batch` already and some docker images require using privileged mode that defaults to `false` when using `local_docker`:
```
docker inspect --format='{{.HostConfig.Privileged}}' <CONTAINER_SHA-1>
false
```

## Motivation/Background
See description

## Detailed Proposal
Add a new option and pass it to Docker client for `local_docker` scheduler.

## Alternatives
Patch `docker_scheduler.py` locally to pass this flag.

## Additional context/links
N/A

PR #898, opened 2024-04-26
feat: add privileged option to local_docker (#897)

clumsy:
Adding `privileged` option for `local_docker` so that the container can start in privileged mode, same way we can do this for `aws_batch` scheduler. Defaults to `False`, same as in `aws_batch`.

Test plan:
☑ unit test added
☑ tested locally:
```
docker inspect --format='{{.HostConfig.Privileged}}' <CONTAINER_ID>
true
```

### DIFF
--- a/torchx/schedulers/docker_scheduler.py
+++ b/torchx/schedulers/docker_scheduler.py
@@ -124,6 +124,7 @@ def ensure_network(client: Optional["DockerClient"] = None) -> None:
 class DockerOpts(TypedDict, total=False):
     copy_env: Optional[List[str]]
     env: Optional[Dict[str, str]]
+    privileged: bool
 
 
 class DockerScheduler(DockerWorkspaceMixin, Scheduler[DockerOpts]):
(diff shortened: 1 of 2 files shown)


clumsy:

clumsy:
Any chance you can check this request, @d4l3k?

clumsy:
Any chance you can look at this one, @andywag?

d4l3k (approved):
LGTM

torchx/schedulers/docker_scheduler.py
```diff
@@ -287,6 +288,7 @@ def _submit_dryrun(self, app: AppDef, cfg: DockerOpts) -> AppDryRunInfo[DockerJo
                             LABEL_REPLICA_ID: str(replica_id),
                         },
                         "hostname": name,
+                        "privileged": cfg["privileged"],
```
d4l3k, line 291: can we change this to be `cfg.get("privileged", False)` for safety?

d4l3k:
Also not sure why it's not giving me an option to run the full tests -- maybe need to bump the PR. 

@clumsy sorry for the delay in reviews on this

torchx/schedulers/docker_scheduler.py
clumsy: Sure, I'll fix that. Just FYI: https://github.com/pytorch/torchx/blob/a2423f4cf9b0a7b3ddad624d8b06ebddf3b8a3cb/torchx/schedulers/aws_batch_scheduler.py#L512

clumsy:
Addressed your comment, @d4l3k. Please let me know if further modifications are needed.

Issue #906, opened 2024-05-06
local_docker does not add utility nvidia libraries to containers


clumsy:
## 🐛 Bug

nvidia Docker images require adding libraries like `libnvidia-ml` that are part of `utility` capability.

TorchX currently only adds `compute` [here](https://github.com/pytorch/torchx/blob/eb7e3d87b10842f2f257b623f5a7ee084630f9f8/torchx/schedulers/docker_scheduler.py#L317).

There are two solutions I verified for this issue, **not sure which one is better**:

* [This PR] add `utility` next to `compute`. Similar fixes [here](https://github.com/portainer/portainer/blob/8a81d95253c1f1946f0ee5bc85db557cc66e602e/app/react/docker/containers/CreateView/ResourcesTab/GpuFieldset/toViewModel.ts#L10) and [here](https://github.com/arvados/arvados/blob/b0079c916db05e728cca119333e6dd6a1afd8a83/lib/crunchrun/docker.go#L172)
* [My personal preference] remove device capabilities from `docker_scheduler` and rely on default values OR let the user customize via `nvidia-container-toolkit`, e.g. we don't set this for `aws_batch_scheduler` and it works fine [here](https://github.com/pytorch/torchx/blob/eb7e3d87b10842f2f257b623f5a7ee084630f9f8/torchx/schedulers/aws_batch_scheduler.py#L236). Not sure if it's backward compatible with old versions of nvidia-container-runtime though.

**NOTE**: [nvidia-container-runtime](https://github.com/NVIDIA/nvidia-container-runtime) has been superseded by [nvidia-container-toolkit](https://github.com/NVIDIA/nvidia-container-toolkit).

Module (check all that applies):
 * [ ] `torchx.spec`
 * [ ] `torchx.component`
 * [ ] `torchx.apps`
 * [ ] `torchx.runtime`
 * [ ] `torchx.cli`
 * [x] `torchx.schedulers`
 * [ ] `torchx.pipelines`
 * [ ] `torchx.aws`
 * [ ] `torchx.examples`
 * [ ] `other`

## To Reproduce

1.
1.
1.

This results in a crash with:
```
train/0 [0]:[rank: 0] Global seed set to 10
train/0 [0]:Initializing distributed: GLOBAL_RANK: 0, MEMBER: 1/1
train/0 [0]:----------------------------------------------------------------------------------------------------
train/0 [0]:distributed_backend=nccl
train/0 [0]:All distributed processes registered. Starting with 1 processes
[...]

PR #907, opened 2024-05-06
fix: add utility capability for local_docker (#906)

clumsy:
nvidia Docker images require adding libraries like `libnvidia-ml` that are part of `utility` capability.

TorchX currently only adds `compute` [here](https://github.com/pytorch/torchx/blob/eb7e3d87b10842f2f257b623f5a7ee084630f9f8/torchx/schedulers/docker_scheduler.py#L317).

Add `utility` next to `compute`. Similar fixes [here](https://github.com/portainer/portainer/blob/8a81d95253c1f1946f0ee5bc85db557cc66e602e/app/react/docker/containers/CreateView/ResourcesTab/GpuFieldset/toViewModel.ts#L10) and [here](https://github.com/arvados/arvados/blob/b0079c916db05e728cca119333e6dd6a1afd8a83/lib/crunchrun/docker.go#L172)

**NOTE**: [nvidia-container-runtime](https://github.com/NVIDIA/nvidia-container-runtime) has been superseded by [nvidia-container-toolkit](https://github.com/NVIDIA/nvidia-container-toolkit).

Test plan:
☑ updated unit test

☑  verified works with `compute,utility`

☑ verfiied works with default device capabilities too

### DIFF
--- a/torchx/schedulers/docker_scheduler.py
+++ b/torchx/schedulers/docker_scheduler.py
@@ -314,7 +314,7 @@ class DockerScheduler(DockerWorkspaceMixin, Scheduler[DockerOpts]):
                     c.kwargs["device_requests"] = [
                         DeviceRequest(
                             count=resource.gpu,
-                            capabilities=[["compute"]],
+                            capabilities=[["compute", "utility"]],
                         )
                     ]
                 req.containers.append(c)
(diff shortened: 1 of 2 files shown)


andywag (approved):
lgtm

clumsy:
Thanks for checking @andywag. Do you happen to know why do we still need device_request capabilities for `local_docker`?

I mentioned an alternative solution in https://github.com/pytorch/torchx/issues/906 to delete this section entirely (e.g. we don't have it for `aws_batch`). Decided to fix in a somewhat backward-compatible fashion here since I'm not aware of the reason this code is there.

PR #911, opened 2024-05-16
Ensure duplicate arguments are only checked within their respective argument groups

hstonec:
Summary:
There are valid scenarios where the same argument may appear in different groups, delineated by '--'. E.g.:

```
torchx run \
--scheduler local_penv \
--workspace //fblearner/flow/projects/lex/ig_o3_torchx:workflow \
-cfg build_remote_package=true  \
fb.python.binary \
--img lex_ig_o3_package  \
-m dper_lib.instagram.pyper_v2.teams.stories.train \
-- \
-m
```
("-m" appears twice in 2 different groups)
Right now, this doesn't pass our duplicate argument checks so this diff changed that.

Differential Revision: D57459718

### DIFF
--- a/torchx/cli/cmd_run.py
+++ b/torchx/cli/cmd_run.py
@@ -13,6 +13,7 @@ import sys
 import threading
 from collections import Counter
 from dataclasses import asdict
+from itertools import groupby
 from pathlib import Path
 from pprint import pformat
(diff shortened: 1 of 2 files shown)


andywag (approved):
lgtm

PR #914, opened 2024-05-20
Adjust CLI dict arg parsing to handle empty string value

manav-a:
Summary: This diff adjusts the dict arg parsing to allow for empty string value to be specified. This makes a trade off the people more often want empty strings than a string containing `""` or `''`. It will be nice to in generally improve the arg parsing setup holistically for torchx in the near future.

Differential Revision: D57515667


andywag (approved):
LGTM

PR #915, opened 2024-05-22
Allow using torchx_ env vars to set scheduler  params

manav-a:
Summary: Scheduler params currently can only be set through the programmatic API and not through this is not useful for cases like scheduling on mast rc cluster. This diff now lets you do that.

Differential Revision: D57640022


andywag (approved):
lgtm!

PR #921, opened 2024-06-14
schedulers: add macro support to metadata variables

jason-b-akers:
Summary:
This adds torchx.spec.macros support to the metadata variable values. This allows use cases like setting smcBridge to the app_id eg:
```
job_spec.roles[0].metadata["mast"] = {
        "HpcTaskGroupSpec": {
            "smcBridge": {
                "portName": "thrift",
                "smcTier": specs.macros.app_id,
            }
        }
    }
```

Reviewed By: manav-a

Differential Revision: D57519702

### DIFF
--- a/torchx/specs/api.py
+++ b/torchx/specs/api.py
@@ -10,6 +10,7 @@
 import copy
 import json
 import re
+import typing
 from dataclasses import asdict, dataclass, field
 from datetime import datetime
 from enum import Enum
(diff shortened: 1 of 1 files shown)


manav-a (approved):

torchx/specs/api.py
```diff
@@ -189,8 +190,25 @@ def apply(self, role: "Role") -> "Role":
             role = copy.deepcopy(role)
             role.args = [self.substitute(arg) for arg in role.args]
             role.env = {key: self.substitute(arg) for key, arg in role.env.items()}
+            role.metadata = self._apply_nested(role.metadata)
```
kiukchung, line 193: could you add a test-case for this?

manav-a: There is a test case on the meta side of this make me can also add a version of it to api test @jason-b-akers

jason-b-akers: Hi @kiukchung. There's an internal test in [D58678835](https://www.internalfb.com/diff/D58678835). As a follow-up I'll write one in torchx/schedulers/test/api_test.py as well.

jason-b-akers: Follow-up test in [PR#293](github.com/pytorch/torchx/pull/923)

PR #922, opened 2024-06-17
Add a new experimental restart policy for large scale model training

manav-a:
Summary: TSIA

Differential Revision: D58684341

### DIFF
--- a/torchx/specs/api.py
+++ b/torchx/specs/api.py
@@ -237,11 +237,15 @@ class RetryPolicy(str, Enum):
                 application to deal with failed replica departures and
                 replacement replica admittance.
     2. APPLICATION: Restarts the entire application.
-
+    3. HOT_SPARE: Restarts the replicas for a role as long as quorum (min_replicas)
+                is not violated using extra hosts as spares. It does not really support
+                elasticity and just uses the delta between num_replicas and min_replicas
+                as spares (EXPERIMENTAL).
     """
 
     REPLICA = "REPLICA"
     APPLICATION = "APPLICATION"
+    HOT_SPARE = "HOT_SPARE"
 
 
 class MountType(str, Enum):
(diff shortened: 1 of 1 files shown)


andywag (approved):
lgtm

torchx/specs/api.py
```diff
@@ -219,11 +219,13 @@ class RetryPolicy(str, Enum):
...
                 replacement replica admittance.
     2. APPLICATION: Restarts the entire application.
-
+    3. QUORUM: Restarts the replicas for a role as long as the quorum is not
+               violated. (EXPERIMENTAL)
     """
 
     REPLICA = "REPLICA"
     APPLICATION = "APPLICATION"
+    QUORUM = "QUORUM"
```
kiukchung, line 228: will the quorum restart policy be checked in? I'm only seeing the specs changes here.

kiukchung (changes requested):

torchx/specs/api.py
manav-a: Its a meta only change to some schedulers there is nothing that has to be handled for existing schedulers

andywag (approved):

torchx/specs/api.py
d4l3k: For existing open source schedulers (i.e. Ray/Volcano) that support some form of elasticity we use `num_replicas` and `min_replicas` fields below to detect elastic behavior

Could we use that as a trigger for this behavior instead?

I'm not totally opposed to adding in a new restart policy but it's unclear what the defined semantics for this is. Currently RetryPolicy defines service vs gang scheduling -- it's not clear that quorum is orthogonal behavior to that.

manav-a: Updated it to APPLICATION_HOT_SPARES just to differentiate between this restart policy and elasticity in general for now. We can see how things evolve on the MAST side and what version of elaticity of replacement we end up with

d4l3k: @kiukchung I think this is generic enough and now well defined enough to be included in the core APIs. Nothing else supports it currently but if anyone uses it it'll just throw an error.

https://github.com/pytorch/torchx/blob/main/torchx/schedulers/kubernetes_scheduler.py#L396

We could throw a better error rather than just KeyError

d4l3k (approved):
LGTM

PR #926, opened 2024-06-26
Use latest pytorch-lightning version

hstonec:
Previously, we had fixed the version to an earlier release, which became incompatible after pip updated to 24.1. So, we took this opportunity to update it to the latest version.

Test plan:
All tests from github

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -20,7 +20,7 @@ pyre-extensions
 pyre-check
 pytest
 pytest-cov
-pytorch-lightning==1.9.0
+pytorch-lightning==1.6.0
 sagemaker>=2.149.0
 torch-model-archiver>=0.4.2
 torch==2.2.1
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -21,6 +21,7 @@ pyre-check
 pytest
 pytest-cov
 pytorch-lightning
+tensorboard
 sagemaker>=2.149.0
 torch-model-archiver>=0.4.2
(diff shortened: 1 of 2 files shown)


dev-requirements.txt
```diff
@@ -20,7 +20,8 @@ pyre-extensions
 pyre-check
 pytest
 pytest-cov
-pytorch-lightning==1.5.10
+pytorch-lightning
+tensorboard
```
yikaiMeta, line 24: why do we need this?

Sanjay-Ganeshan:
Instead of using the latest version of pytorch lightning - can we set the specific version that happens to be the latest at the moment, with an `==` of `>=`?

That's 2.3.1, right now

https://pypi.org/project/pytorch-lightning/#history

Otherwise, LGTM

Sanjay-Ganeshan (approved):
Consider setting `pytorch-lightning==2.3.1`

Otherwise, LGTM

dev-requirements.txt
hstonec: The latest pytorch-lightning needs that: 

image

hstonec:
> Instead of using the latest version of pytorch lightning - can we set the specific version that happens to be the latest at the moment, with an `==` of `>=`?
> 
> That's 2.3.1, right now
> 
> https://pypi.org/project/pytorch-lightning/#history
> 
> Otherwise, LGTM

Yeah, we can do that if that can reduce some maintenance work in the future.

PR #933, opened 2024-07-17
Add functionality to load scheduler cfg from a json string

ishachirimar:
Summary:
add a function `cfg_from_json_repr()` to create scheduler runopts from a json string representation.

currently, there is only `cfg_from_str()` which takes a string in the format "k=v,k1=v1" (designed for cmd line). however, the runcfg is stored in scuba logs in a json string representation.

so in order to read it back when cloning jobs we need to read from json string.

Differential Revision: D59832084


andywag (approved):

Sanjay-Ganeshan (approved):

PR #935, opened 2024-07-19
Allow field path env variables

ndacosta:
Summary:
# Context
Currently the kubernetes scheduler translate all the env of the role into something like this.
```
- name: FOO
  value: bar
```
In some cases we may also need env variables of type:
```
- name: FOO
  valueFrom:
    fieldRef:
      fieldPath: bar.path
```
This diff adds the ability to specify env in the role that will be translated to the second option.

# Changes
- Create constant placeholder to insert in the value of env in roles if we want them to be translated to the second option
- If placeholder is present in the value, convert to code below stripping the placeholder
```
V1EnvVar(
  name=key,
    value_from=V1EnvVarSource(
      field_ref=V1ObjectFieldSelector(field_path=value)
    ),
)
```
if not, translate to
```
V1EnvVar(name=key, value=value)
```
- update tests

Differential Revision: D59865500

### DIFF
--- a/torchx/schedulers/kubernetes_scheduler.py
+++ b/torchx/schedulers/kubernetes_scheduler.py
@@ -169,6 +169,17 @@ ANNOTATION_ISTIO_SIDECAR = "sidecar.istio.io/inject"
 
 LABEL_INSTANCE_TYPE = "node.kubernetes.io/instance-type"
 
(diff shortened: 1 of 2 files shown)


Sanjay-Ganeshan (approved):
LGTM - let's add a comment above the constant value

PR #937, opened 2024-07-22
Sync pyre strict comments to match internal copy

Sanjay-Ganeshan:
The internal version of torchx has pyre-strict comments. The external version does not.
This diff adds the corresponding comments to re-sync the two repos, using the internal version as a the source of truth.

Test plan:
(Comments - no changes to behaviour, except for pyre)


yikaiMeta (approved):
thanks for fixing this

PR #939, opened 2024-07-26
Refactor kubernetes scheduler to have a base scheduler

ndacosta:
Summary:
# Context
Refactoring the kubernetes scheduler so we can have different implementations with different workspaces but we can still reuse the common functionality.

# Changes
- Extract kubernetes specific functionality into a BaseKubernetesScheduler
- make KubernetesScheduler inherit from BaseKubernetesScheduler

Differential Revision: D60174131


kiukchung:
The corresponding internal change (D60174131) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #947, opened 2024-08-21
Add appdef metadata to torchx event

zackycao:
Summary: Add appdef metadata to torchx event class. Scuba logs does not get modified yet.

Differential Revision: D61632621


andywag (approved):

PR #950, opened 2024-08-26
Fix python-unittest.yaml

yikaiMeta:
Test plan:


yikaiMeta:
the slurm failure is a pre-existing issue.

d4l3k (approved):

PR #953, opened 2024-09-03
Create session id and use it in table 'Pytorch Elastic Tsm Log'

yikaiMeta:
We need to create a unique session ID to identify each torchx run session.

Differential Revision: D62087199


andywag (approved):

PR #956, opened 2024-09-16
Adding Override Option for TorchX Role

andywag:
Summary: Adds a generic way to override the internal values of the Role. Allows async overriding of role values and enable Async Packaging

Differential Revision: D62591176


Sanjay-Ganeshan (approved):
LGTM - same as internal version

PR #959, opened 2024-09-30
expose torchx session ID to torchx downstream users.

yikaiMeta:
Summary: this diff creates a function which let customers read the torchx session ID.

Differential Revision: D63655049


andywag (approved):

PR #965, opened 2024-10-14
Standalone Workspace build for TorchX

andywag:
Summary:
Option to build workspace independently of scheduler with async option.

First Cut to get agreement oninterface

Differential Revision: D63145971


Sanjay-Ganeshan (approved):
LGTM - the second class is necessary because "default" typing doesn't work in the relevant versions of Python

PR #973, opened 2024-11-20
Strip lines when log-tailing

Sanjay-Ganeshan:
Summary:
For some reason, certain jobs have logs that end in \n\n, making unreadable whitespace

This diff explicitly strips lines before prefixing them, and then manually adds back exactly one "\n"

Created from CodeHub with https://fburl.com/edit-in-codehub

Reviewed By: haonans

Differential Revision: D66191123


tonykao8080 (approved):
looks good

PR #974, opened 2024-11-22
Add torchx session id as Environment variable

yikaiMeta:
Summary:
This diff passes the torchx_session_id as an environment variable.

In penv_python, I will fetch this id and log it into logging table.
With the session id, it is possible to join the logging table of penv_python with the logging table of torchx.

Differential Revision: D66387522


andywag (approved):
lgtm

PR #975, opened 2024-11-25
Update .pyre_configuration to fix Pyre

yikaiMeta:
Test plan:


tonykao8080 (approved):
approve to unblock lint

PR #981, opened 2024-12-03
Fix pyre errors

jesszzzz:
Title says it all

Test plan:
Ran `scripts/pyre.sh`
Also waiting for CI


kiukchung:
Thanks! These pyre fixes have drifted - the type errors here were since addressed on main and the type-checking setup has moved on (pyre to pyrefly migration). Closing; please re-spin against current main if anything still flags.

PR #984, opened 2024-12-06
Fix docs build by freezing apache airflow version

jesszzzz:
apache-airflow v2.10.3 bumped a dependency that broke kfp protobuf stuff in docs, this freezes the version to stick to a compatible version

Test plan:

Docs-related CI passes (other failures are pre-existing)


kiukchung:
Thanks! The docs build has since moved to uv-managed pinned dependencies and apache-airflow/kfp are no longer part of the docs requirements, so this pin no longer applies. Closing - please re-open if you hit a docs-build issue on current main.

PR #986, opened 2024-12-09
sync torchx .pyre_configuration.internal with external config and upg…

stroxler:
There are some issues with Pyre versions in github CI, and at the moment we aren't getting type errors.

Making a PR to
(a) upgrade to the  2024-11-25 release
(b) suppress errors

I'm making a fresh PR because the original one doesn't seem to trigger github CI (maybe because it was exported from a diff)


jesszzzz (approved):
Amazing, thank you!!

PR #1006, opened 2025-02-07
Fix for issue with APF packaging with multiple roles

andywag:
Summary: There is an issue with the async logic when there are multiple jobs/roles associated with single package. This fix solves this by allowing the future to be used by more than one arm of the package.

Reviewed By: ishachirimar, chouxi

Differential Revision: D69279705


tonykao8080 (approved):
looks good

ishachirimar (approved):

PR #1007, opened 2025-02-11
torchx support early validation before workspace build

tonykao8080:
Summary:
add `_pre_build_validate()` hook for torchx scheduler to perform app validation before build workspace step. This earlier validation enables detecting issue sooner without the need to wait for build workspace to complete. This change only exposes the pre_build_validate() hook for torchx scheduler and there is no change to existing behavior (validation will continue to perform after build workspace if required). Subsequent change will change the behavior within specific scheduler if validation can be moved from validate to pre_build_validate.

additional change to add event logging for build_workspace_and_update_role call.

Differential Revision: D69463377

### DIFF
--- a/torchx/runner/api.py
+++ b/torchx/runner/api.py
@@ -412,6 +412,14 @@ class Runner:
         ):
             sched = self._scheduler(scheduler)
             resolved_cfg = sched.run_opts().resolve(cfg)
+
+            # early validation before build workspace
+            with log_event(
+                "pre_build_validate",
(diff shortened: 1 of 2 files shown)


andywag (approved):
lgtm

lgarg26 (approved):

ishachirimar (approved):
lgtm - some of the schedulers log from within `build_workspace_and_update_role` but doesn't look like a concern for duplication

PR #1010, opened 2025-02-20
Allow running shell script from fbpkg

hstonec:
Summary: Previously `sh` is hard coded to be the entrypoint which doesn't work in MAST, so this diff allows to use a user provided entry point.

Differential Revision: D69951935


andywag (approved):

PR #1011, opened 2025-02-21
Handle args for customized entrypoint

hstonec:
Summary: When customized entrypoint is provided, the args need to be persevered in the original format.

Differential Revision: D69964149


andywag (approved):
lgtm

PR #1020, opened 2025-03-20
Drop python-3.8 unittest in favor of adding 3.12. Update to python-3.10 for other workflows

kiukchung:
Summary:
PyTorch dropped support for python-3.8 a while ago. This change makes TorchX consistent with PyTorch. This should also fix gh workflow breakages in https://hud.pytorch.org/hud/pytorch/torchx/main

Changes:

1. Drops python-3.8 in unittest matrix
2. Adds python-3.12 to unittest matrix
3. Upgrades a few libraries (torch, torchvision, etc) in dev-requirements.txt
4. Upgrades to python-3.10 in non-unittest workflows

Reviewed By: manav-a

Differential Revision: D71475986

### DIFF
--- a/.github/workflows/aws-batch-integration-tests.yaml
+++ b/.github/workflows/aws-batch-integration-tests.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.9
+          python-version: "3.10"
           architecture: x64
       - name: Checkout TorchX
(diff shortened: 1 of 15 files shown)


d4l3k (approved):
LGTM

There some other changes as well -- is that intended?

kiukchung:
> LGTM
> 
> There some other changes as well -- is that intended?

yea, the other changes (ones in *.py files) are needed to run tests in python-3.12.

PR #1022, opened 2025-03-21
schedulers parameterized

bobyangyf:
Summary:
Parameterize Scheduler
This not only more precisely expresses it's type (previously was unspecified generic variant of AppDryRunInfo[T] where T was unspecified; not it can be fully specified); This also allows us to inject PipelineDryRunInfo for schedulers that are pipeline schedulers

Differential Revision: D71023681

### DIFF
--- a/torchx/schedulers/api.py
+++ b/torchx/schedulers/api.py
@@ -94,9 +94,11 @@ class ListAppResponse:
 
 
 T = TypeVar("T")
+A = TypeVar("A")
+D = TypeVar("D")
 
 
-class Scheduler(abc.ABC, Generic[T]):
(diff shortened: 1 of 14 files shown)


kiukchung (commented):
I don't fully understand this change. It seems like you want `torchx.schedulers.api.Scheduler` interface to be able to deal with both `AppDef` and `PipelineDef` in the **same** API. For instance:

```
def schedule(desc: Union[AppDef, PipelineDef], ...)
```

If this is the case, I would discourage you from overloading the scheduler APIs in such a way. 

I'd have to fully understand what the intent here is before recommending a good path forward. Happy to meet and discuss.

lgarg26 (approved):
Accepting to unblock ongoing intergration.

PR #1029, opened 2025-03-26
Fix broken component integration test due to compute_world_size app not respecting env vars set by torchrun

kiukchung:
Summary:
`compute_world_size` is run as an integration test in a `-j 2x2` configuration using `torchrun` which sets `MASTER_ADDR` and `MASTER_PORT`. However, it was ignoring those env vars and overriding them with the ones in the hydra config (added to make `compute_world_size` work as a single process without `torchrun`).

Integ tests are failing in CI because `localhost:0` (pick random free port) is used as the `MASTER_ADDR:MASTER_PORT` on all 4 workers, hence all 4 workers are deadlocked waiting for each other to join the job.

This diff fixes this by only setting the env vars if one is not already set.

Differential Revision: D71919903


tonykao8080 (approved):
looks good

PR #1032, opened 2025-03-27
Run CI actions on ubuntu 24.04 since ubuntu 20.04 is being deprecated on 04/01/2025

kiukchung:
Summary:
We've been getting a bunch of warnings in our CI because ubuntu 20.04 is being deprecated in a few days. This change upgrades us to ubuntu 24.04

 {F1976524616}

Differential Revision: D72004581


tonykao8080 (approved):
lgtm

PR #1034, opened 2025-03-28
Fix broken doc-push

kiukchung:
Summary:
I fixed doc-build in https://github.com/pytorch/torchx/pull/1033 by:

1. pinning `protobuf-3.20.x` in torchx's `dev-requirements.txt` since `apache-airflow` (defined in `docs/requirements.txt`) was upgrading to `protobuf>=5.x` which is incompatible with `kfp-1.8.x`.
2. installing `doc/requirements.txt` FIRST, then installing `torchx` so that the pinned version of `protobuf` is honored.

The above fixes `doc-build` but not `doc-push` because `doc-push` runs `scripts/doc-push.sh` which needed the same change as #2.

Differential Revision: D72059926

### DIFF
--- a/docs/doc_push.sh
+++ b/docs/doc_push.sh
@@ -54,8 +54,12 @@ else
     redirects=(latest main)
 fi
 
-echo "Installing torchx from $repo_root..."
+echo "Installing TorchX and Doc dependencies from $repo_root..."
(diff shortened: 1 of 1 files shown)


tonykao8080 (approved):
looks good

PR #1039, opened 2025-03-31
[torchx][CI] Run k8s integration test on linux.24_04.4x since the *.1…

kiukchung:
k8s integ tests are failing because `linux.24_04.16x` takes too long to acquire. Switch over to a smaller instance (`linux.24_04.4x`).


tonykao8080 (approved):
looks good

PR #1043, opened 2025-04-11
flag to log full trace on error

tonykao8080:
Summary:
While trying to debug an asyncio RuntimeError, I realized that the current exception logging in log_event does not include the full stacktrace of the error (it only starts from where the error was thrown).
Adding the ability to include the full stack to help with debugging

Differential Revision: D72800297


kiukchung:
The corresponding internal change (D72800297) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

Issue #1009, opened 2025-02-17
List all registered torchx.schedulers


clumsy:
## Description
Currently we can only have the default ones or the custom ones (from one of the packages): https://github.com/pytorch/torchx/blob/main/torchx/schedulers/__init__.py#L59

## Motivation/Background
This will allow adding new schedulers locally without re-registering the default schedulers. The list gets stale when new default schedulers are added to TorchX

## Detailed Proposal
Make it work the same way as named_resources (simply list all) or components (allow referencing via _0, _1, etc.)

## Alternatives
Keep re-registering the default schedulers

## Additional context/links
* https://github.com/pytorch/torchx/blob/main/torchx/schedulers/__init__.py#L59
* https://github.com/pytorch/torchx/blob/main/torchx/specs/__init__.py#L65
* https://github.com/pytorch/torchx/blob/420921847296790e8bfa5cd43c8d9776a3f88165/torchx/specs/finder.py#L322

clumsy:
For your consideration, @d4l3k @kiukchung @tonykao8080

clumsy:
The fix to simply merge/overwrite existing schedulers is trivial (publishing PR). The only question is whether we want to support the same `_0`/`_1` behavior as for components [here](https://pytorch.org/torchx/main/advanced.html#registering-custom-components)

PR #1050, opened 2025-04-23
feat: list all registered schedulers (#1009)

clumsy:
A simple merge for the list of all registered schedulers.

Test plan:
[x] all existing tests should pass

### DIFF
--- a/torchx/schedulers/__init__.py
+++ b/torchx/schedulers/__init__.py
@@ -13,7 +13,7 @@ from typing import Mapping, Protocol
 from torchx.schedulers.api import Scheduler
 from torchx.util.entrypoints import load_group
 
-DEFAULT_SCHEDULER_MODULES: Mapping[str, str] = {
+BUILTIN_SCHEDULER_MODULES: Mapping[str, str] = {
     "local_docker": "torchx.schedulers.docker_scheduler",
     "local_cwd": "torchx.schedulers.local_scheduler",
     "slurm": "torchx.schedulers.slurm_scheduler",
@@ -39,8 +39,17 @@ def _defer_load_scheduler(path: str) -> SchedulerFactory:
     return run
 
 
+def default_schedulers() -> dict[str, SchedulerFactory]:
+    """Build default schedulers (built-in + extras from torchx.schedulers.extra)."""
+    return {
+        **{s: _defer_load_scheduler(p) for s, p in BUILTIN_SCHEDULER_MODULES.items()},
+        **load_group("torchx.schedulers.extra", default={}),
+    }
+
+
 def get_scheduler_factories(
-    group: str = "torchx.schedulers", skip_defaults: bool = False
+    group: str = "torchx.schedulers",
+    skip_defaults: bool = False,
 ) -> dict[str, SchedulerFactory]:
     """
     get_scheduler_factories returns all the available schedulers names under `group` and the
@@ -48,15 +57,7 @@ def get_scheduler_factories(
 
     The first scheduler in the dictionary is used as the default scheduler.
     """
-
-    if skip_defaults:
-        default_schedulers = {}
-    else:
-        default_schedulers: dict[str, SchedulerFactory] = {}
-        for scheduler, path in DEFAULT_SCHEDULER_MODULES.items():
-            default_schedulers[scheduler] = _defer_load_scheduler(path)
-
-    return load_group(group, default=default_schedulers)
+    return load_group(group, default={} if skip_defaults else default_schedulers())
 
 
 def get_default_scheduler_name() -> str:
--- a/torchx/schedulers/test/registry_test.py
+++ b/torchx/schedulers/test/registry_test.py
@@ -43,3 +43,75 @@ class SchedulersTest(unittest.TestCase):
 
         for scheduler in schedulers.values():
             self.assertEqual("test_session", scheduler.session_name)
+
+    @patch("torchx.schedulers.load_group")
+    def test_torchx_schedulers_overrides_all(self, mock_load_group: MagicMock) -> None:
+        """torchx.schedulers completely overrides defaults and ignores extras"""
+        mock_custom: MagicMock = MagicMock()
+        mock_extra: MagicMock = MagicMock()
+
+        mock_load_group.side_effect = lambda group, default: (
+            {"custom": mock_custom}
+            if group == "torchx.schedulers"
+            else {"extra": mock_extra} if group == "torchx.schedulers.extra" else {}
+        )
+
+        factories = get_scheduler_factories()
+
+        self.assertEqual(factories, {"custom": mock_custom})
+        self.assertNotIn("local_docker", factories)
+        self.assertNotIn("extra", factories)
+
+    @patch("torchx.schedulers.load_group")
+    def test_no_custom_returns_defaults_and_extras(
+        self, mock_load_group: MagicMock
+    ) -> None:
+        """no custom schedulers returns built-in + extras"""
+        mock_extra: MagicMock = MagicMock()
+
+        mock_load_group.side_effect = lambda group, default: (
+            {"extra": mock_extra} if group == "torchx.schedulers.extra" else default
+        )
+
+        factories = get_scheduler_factories()
+
+        self.assertIn("local_docker", factories)
+        self.assertIn("slurm", factories)
+        self.assertIn("extra", factories)
+
+    @patch("torchx.schedulers.load_group")
+    def test_no_custom_no_extras_returns_builtins(
+        self, mock_load_group: MagicMock
+    ) -> None:
+        """no custom, no extras returns only built-in schedulers"""
+        mock_load_group.side_effect = lambda group, default: default
+
+        factories = get_scheduler_factories()
+
+        self.assertIn("local_docker", factories)
+        self.assertIn("slurm", factories)
+
+    @patch("torchx.schedulers.load_group")
+    def test_skip_defaults_returns_empty(self, mock_load_group: MagicMock) -> None:
+        """skip_defaults=True with no custom schedulers returns empty"""
+        mock_load_group.side_effect = lambda group, default: default
+
+        factories = get_scheduler_factories(skip_defaults=True)
+
+        self.assertEqual(factories, {})
+
+    @patch("torchx.schedulers.load_group")
+    def test_custom_scheduler_is_default(self, mock_load_group: MagicMock) -> None:
+        """first custom scheduler becomes the default"""
+        mock_aws: MagicMock = MagicMock()
+        mock_custom: MagicMock = MagicMock()
+
+        mock_load_group.side_effect = lambda group, default: (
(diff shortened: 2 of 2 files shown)


kiukchung:
could you provide more context to what you want to achieve with this?

clumsy:
All the details are in the linked #1050, @kiukchung. Please let me know if more details are needed there.

kiukchung:
Hi @clumsy thanks for the pointer. `torchx.schedulers.get_scheduler_factories()` is a public API and this change is not backwards-compatible for the case `get_scheduler_factories(skip_defaults=False)` where there exists registered entrypoint schedulers. Now users will get their configured + default schedulers instead of just their configured ones.

Could you describe your use-case in wanting the list of supported schedulers offered to your users to be dynamic? Usually torchx users want to control the schedulers they configure for their users.

clumsy:
Sure, @kiukchung

Take NeMo for example, NVidia ships it with all dependencies, including `nemo-run` (https://github.com/NVIDIA/NeMo/blob/94589bde88fab1997c842be4e000faf69180cffb/nemo/collections/common/parts/nemo_run_utils.py#L18)

Unfortunately nemo-run unconditionally registers custom schedulers: https://github.com/NVIDIA/NeMo-Run/blob/main/pyproject.toml#L43-L48

Thus we cannot use `local_cwd` from within the container for example, or if we have nemo-run installed.

It makes sense to have a feature to restrict the available schedulers, but does it have to be the default one?

kiukchung:
@clumsy ah that's an interesting edge-case. What you basically want is for `torchx.schedulers` to be additive. We don't treat it as such today. Since Python entrypoint groups don't compound, we have to come up with a convention for the group names. 

We could do something like: `{org_name}.torchx.schedulers` and at load time select `*.torchx.schedulers` entrypoint groups. For BC we'd also have to keep reading `torchx.schedulers`.

If you're open to it, you can add support for prefixes in `torch.util.entrypoints.load()` (https://github.com/pytorch/torchx/blob/9120355959fef8eb438226e9521a07a00e02078e/torchx/util/entrypoints.py#L54)

and make a change in `torchx.schedulers.get_schedulers()` to call the `load()` fn appropriately.

There's some interesting cases regarding name conflicts and ordering (e.g. if nemo registers a scheduler with the same name as the one somewhere else what do you do?)

clumsy:
Thanks, @kiukchung! The proposed solution works for me and I can provide the implementation shortly.

clumsy:
Please check this implementation, @kiukchung , @andywag, @d4l3k, @tonykao8080 

The existing commands continue to work, e.g.:
* `torchx runopts`
```
my_package.custom_local_docker:
    usage:
        [copy_env=COPY_ENV],[env=ENV],[privileged=PRIVILEGED],[debug=DEBUG],[image_repo=IMAGE_REPO],[quiet=QUIET]

    optional arguments:
        copy_env=COPY_ENV (typing.List[str], None)
            list of glob patterns of environment variables to copy if not set in AppDef. Ex: FOO_*
        env=ENV (typing.Dict[str, str], None)
            environment variables to be passed to the run. The separator sign can be eiher comma or semicolon
            (e.g. ENV1:v1,ENV2:v2,ENV3:v3 or ENV1:V1;ENV2:V2). Environment variables from env will be applied on top
            of the ones from copy_env
        privileged=PRIVILEGED (bool, False)
            If true runs the container with elevated permissions. Equivalent to running with `docker run --privileged`.
        debug=DEBUG (bool, False)
            run a container with noop entrypoint, useful for debugging environment
        image_repo=IMAGE_REPO (str, None)
            (remote jobs) the image repository to use when pushing patched images, must have push access. Ex: example.com/your/container
        quiet=QUIET (bool, False)
            whether to suppress verbose output for image building. Defaults to ``False``.
```

* `torchx run -s my_package.custom_local_cwd`
```
usage: torchx run [-h]
                  [-s {local_docker,local_cwd,slurm,kubernetes,kubernetes_mcad,aws_batch,aws_sagemaker,gcp_batch,ray,lsf,my_package.custom_aws_batch,my_package.aws_sagemaker,my_package.custom_gcp_batch,my_package.kubernetes,my_package.kubernetes_mcad,my_package.custom_local_cwd,my_package.local_docker,my_package.lsf,my_package.custom_aws_batch,my_package.custom_local_docker,}]
```

* `torchx run -s my_package.custom_local_cwd utils.echo --msg "test"`
```
torchx 2025-05-13 12:52:15 INFO     Tracker configurations: {}
torchx 2025-05-13 12:52:15 INFO     Log directory not set in scheduler cfg. Creating a temporary log dir that will be deleted on exit. To preserve log directory set the `log_dir` cfg option
torchx 2025-05-13 12:52:15 INFO     Log directory is: /var/folders/8b/nbn0wcb93m710myrqx8r_clh0000gq/T/torchx_m0wx4hxg
my_package.custom_local_cwd://torchx/echo-gww0rzcc0z72tc
torchx 2025-05-13 12:52:15 INFO     Launched app: my_package.custom_local_cwd://torchx/echo-gww0rzcc0z72tc
torchx 2025-05-13 12:52:15 INFO     AppStatus:
    State: RUNNING
    Num Restarts: 0
    Roles:
    Msg: <NONE>
    Structured Error Msg: <NONE>
    UI URL: file:///var/folders/8b/nbn0wcb93m710myrqx8r_clh0000gq/T/torchx_m0wx4hxg/torchx/echo-gww0rzcc0z72tc
```

clumsy:
The failure seems unrelated to the PR: ``` E   RuntimeError: operator torchvision::nms does not exist```
@kiukchung Should we rerun the tests?

clumsy:
Hi @kiukchung, should I do a bogus push to trigger new test run? The last failure seem to have been unrelated.

clumsy:
Hi @kiukchung does this PR need more work or is there anything I can do to help integrate the change?

clumsy:
Is there anything I can do here, @kiukchung?

torchx/schedulers/__init__.py
```diff
@@ -48,15 +48,16 @@ def get_scheduler_factories(
...
-    return load_group(group, default=default_schedulers)
+    valid_schedulers = (
+        {}
+        if skip_defaults
+        else {
+            name: _defer_load_scheduler(path)
+            for name, path in DEFAULT_SCHEDULER_MODULES.items()
+        }
+    )
+    valid_schedulers.update(load_group(group, default={}))
```
kiukchung: this makes `get_scheduler_factories()` non BC no?

For example say I had `DEFAULT__SCHEDULER_MODULES` have `[k8s, slurm]` and in my entry-points.txt I have
```
[torchx.schedulers]
aws_batch = ...
custom_1 = ...
```

prior to this PR:

```
get_scheduler_factories(skip_defaults=False) == {"aws_batch", "custom_1"}
get_scheduler_factories(skip_defaults=True)  == {"aws_batch", "custom_1"}
```
and if I didn't have `[torchx.schedulers]` in entry-points
```
get_scheduler_factories(skip_defaults=False) == {"k8s", "slurm"}
get_scheduler_factories(skip_defaults=True)  == {}
```

Now after this PR

If I had `[torchx.schedulers]` defined in entry-points
```
get_scheduler_factories(skip_defaults=False) == {"k8s", "slurm", "aws_batch", "custom_1"}
get_scheduler_factories(skip_defaults=True)  == {"aws_batch", "custom_1"}
```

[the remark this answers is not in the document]
clumsy: Fixed in new revision and added more unit tests

torchx/schedulers/__init__.py
```diff
@@ -48,15 +48,12 @@ def get_scheduler_factories(
...
-        for scheduler, path in DEFAULT_SCHEDULER_MODULES.items():
-            default_schedulers[scheduler] = _defer_load_scheduler(path)
-
-    return load_group(group, default=default_schedulers)
+    schedulers = load_group(group, default={})
+    if not skip_defaults:
+        for name, path in DEFAULT_SCHEDULER_MODULES.items():
+            if name not in schedulers:
+                schedulers[name] = _defer_load_scheduler(path)
+    return schedulers
```
kiukchung: thanks for addressing one of the BC incompatibilities. But I think the other one still remains. `get_scheduler_factories(skip_defaults=False)` now returns BOTH my custom `[torchx.schedulers]` mappings AND the `DEFAULT_SCHEDULER_MODULES` which is not ideal for users who expect ONLY my custom schedulers to show. This is important since the default schedulers may not even be available for the users to use.

Unlike  named_resources, scheduler_factories is not additive. We should've kept the same behavior in both (either additive or not).

`skip_defaults` argument in `get_scheduler_factories()` means "even if you don't find anything in entry-points, do not return schedulers in `DEFAULT_SCHEDULER_MODULES`. I think the confusion is that `DEFAULT_SCHEDULER_MODULES` is a badly worded constant. What it "should've been called are `BUILTIN_SCHEDULER_MODULES` (aka the ones that torchx has "builtin" support for) and `skip_defaults` should've been called `skip_builtins`.

If the motivation here is for the CLI to expose custom + builtins such that builtins is always in-sync with any new schedulers added to torchx, we are currently in a mode where we are retiring (not adding) schedulers. The goal being better support/integration rather than wider coverage.

clumsy: The goal of this PR is to be able to allow registering custom schedulers next to built-in ones. E.g. if NeMo wants to register some new ones or even overwrite _some_ of them - the rest should be available, e.g. `local_cwd`. It's not practical to ask the user to adding all built-ins manually - we are not even sure in what order these will be resolved, e.g. both NeMo and some other package registering their own. The simplest idea was to allow additive behavior if we are not using the same names. I guess we can preserve the existing "clear+add" approach as well @kiukchung

tonykao8080 (approved):

kiukchung:
> Could you please import the recent revision, @kiukchung @tonykao8080?

this fails internal tests because of the non-BC issue we've discussed. `get_scheduler_factories()` returns either the schedulers defined in `[torchx.schedulers]` entrypoint OR if no such entrypoint group, returns the set of DEFAULT_SCHEDULERS.

Since we have `[torchx.schedulers]` entrypoint set internally, we don't expect the default set of schedulers to get loaded. Especially since we don't actually include dependencies to schedulers like `"kubernetes"` (which we don't use internally).

There are call-sites where users iterate through the loaded scheduler factories and actually create scheduler instances. This PR makes it such that schedulers such as `kubernetes`, `kubernetes_mcad` attempted to be created, which fails since we don't include those modules in our internal deployment of torchx.

kiukchung (changes requested):
See my comment. This change may not be necessary to achieve what you want.

clumsy:
Hi @kiukchung had a chance to test with bogus pyproject.toml with empty `[torchx.schedulers]` - that won't add default schedulers unless we re-register them. What should we settle with?

kiukchung:
The corresponding internal change (D74743585) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

Issue #1052, opened 2025-04-24
quoting support for to_dict


clumsy:
## Description
`to_dict` used to parse `env` key-value pairs does not allow using `=` as part of the value.

## Motivation/Background
This means we cannot set values for `NCCL_IB_HCA` for example: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#id8

## Detailed Proposal
In addition to existing functionality add support for quoted values to escape all special symbols: `;`, `,` and `=`

## Alternatives
Hard code the env variables in, say, Dockerfile

## Additional context/links
[](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#id8)

clumsy:
For you consideration @kiukchung, @d4l3k, @tonykao8080

clumsy:
Merged

PR #1049, opened 2025-04-23
fix: ray module not found handling

clumsy:
TorchX has been handling `ModuleNotFoundError` gracefully for a while now, e.g. for SageMaker when running `torchx runopts` we get:

```
...
            (remote jobs) the image repository to use when pushing patched images, must have push access. Ex: example.com/your/container
        quiet=QUIET (bool, False)
            whether to suppress verbose output for image building. Defaults to ``False``.

aws_sagemaker: No module named 'sagemaker'

gcp_batch:
    usage:
        [project=PROJECT],[location=LOCATION]
...
```

But for `ray` we get an exception after which we won't get next runopts:
```
gcp_batch:
    usage:
        [project=PROJECT],[location=LOCATION]

    optional arguments:
        project=PROJECT (str, None)
            Name of the GCP project. Defaults to the configured GCP project in the environment
        location=LOCATION (str, us-central1)
            Name of the location to schedule the job in. Defaults to us-central1

Traceback (most recent call last):
  File "/usr/local/bin/torchx", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/torchx/cli/main.py", line 118, in main
    run_main(get_sub_cmds(), argv)
  File "/usr/local/lib/python3.10/dist-packages/torchx/cli/main.py", line 114, in run_main
    args.func(args)
  File "/usr/local/lib/python3.10/dist-packages/torchx/cli/cmd_runopts.py", line 36, in run
    opts = runner.scheduler_run_opts(scheduler)
  File "/usr/local/lib/python3.10/dist-packages/torchx/runner/api.py", line 473, in scheduler_run_opts
    return self._scheduler(scheduler).run_opts()
  File "/usr/local/lib/python3.10/dist-packages/torchx/runner/api.py", line 718, in _scheduler
    sched = factory(self._name, **self._scheduler_params)
  File "/usr/local/lib/python3.10/dist-packages/torchx/schedulers/__init__.py", line 39, in run
    module = importlib.import_module(path)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
[...]

PR #1053, opened 2025-04-24
feat: add quoting support to to_dict (#1052)

clumsy:
See #1052

Test plan:
[x] all existing test cases pass

[x] add new test case

[x] test locally:

Given:
```
[component:dist.ddp]
script=test.py
env=
	NCCL_IB_HCA='=mlx5_0:1,mlx5_1:1';
```

Observe:
```
test/0 [7]:environ({...'NCCL_IB_HCA': '=mlx5_0:1,mlx5_1:1', ...})
```

### DIFF
--- a/torchx/util/test/types_test.py
+++ b/torchx/util/test/types_test.py
@@ -197,6 +197,11 @@ class TypesTest(unittest.TestCase):
             to_dict("key1=value1,,foo=bar"),
         )
 
+        self.assertDictEqual(
+            {"FOO": "value with = and , and ;"},
+            to_dict('FOO="value with = and , and ;"'),
+        )
+
     def test_to_dict_malformed_literal(self) -> None:
         for malformed in ["FOO", "FOO,", "FOO;", "FOO=", "FOO=;BAR=v1"]:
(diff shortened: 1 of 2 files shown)


clumsy:
Please consider this PR, @kiukchung, @d4l3k, @tonykao8080 and sorry for using regex 🙂

clumsy:
No particular preference besides being convenient to see the exact value without unescaping, @andywag - would be happy with any working solution.

Do you think an alternative mechanism is more desirable? Do you mean escaping like `\=`?

andywag:
Yeah, I prefer the more standard escape like you mention but am fine with this since it's working.

andywag (approved):
lgtm

clumsy:
Addressed your comment in https://github.com/pytorch/torchx/pull/1049#issuecomment-2833817150 @andywag 

Please let me know if you prefer escaping after all, I have no preference but want to follow the principle of least astonishment.

andywag:
@clumsy Can you add the type to quoted_string quoted_values: List[str] = []? to fix a pyre issue

clumsy:
Done, @andywag

PR #1056, opened 2025-04-28
fix __file__ value inside components

daniel-ohayon:
Summary:
Currently when we exec a component, we pass it the caller file's namespace. As a result, `__file__` resolves to `.../torchx/specs/finder.py` inside the compoennt at runtime, which is confusing.
This fixes that.

Differential Revision: D73793210

### DIFF
--- a/torchx/specs/finder.py
+++ b/torchx/specs/finder.py
@@ -7,6 +7,7 @@
 # pyre-strict
 
 import abc
+import copy
 import importlib
 import inspect
 import logging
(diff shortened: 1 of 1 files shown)


torchx/specs/finder.py
```diff
@@ -281,7 +282,9 @@ def find(
         )
 
         file_source = read_conf_file(self._filepath)
-        namespace = globals()
+        namespace = copy.copy(globals())
+        # so that __file__ used inside the component points to the correct file
+        namespace["__file__"] = os.path.abspath(self._filepath)
```
kiukchung, line 287: would be great to add a unittest assertion for this (either as a testcase or simply adding an assertion in an existing testcase here: https://github.com/pytorch/torchx/blob/main/torchx/specs/test/finder_test.py#L194)

kiukchung (approved):

PR #1062, opened 2025-05-05
(torchx/local_scheduler) Use os.kill instead of os.killpg when sending SIGTERM to the replica pid. Add runner.wait() for torchx.runner.test.api_test#test_empty_session_id to gracefully wait for the replicas to finish running

kiukchung:
Summary:
Fixes macos unittest failures: https://github.com/pytorch/torchx/actions/runs/14844253481/job/41674243877

When looking into the test failure I noticed two things:

1. `local_scheduler` was trying to SIGTERM the process group by passing the replica's pid:  `os.killpg(replica.pid, signal.SIGTERM)` . Changed to call `os.kill`. (note that `os.killpg` is not available on iOS which is why the test was failing).

2. The `torchx.runner.test.api_test.test_empty_session_id()` test case doesn't wait for the `echo` test command to finish hence there was a race condition where in certain cases the runner's `__exit__()` SIGTERMs the replica pids but since the `local_scheduler` was (wronfully) using `os.killpg` not `os.kill` it threw an uncaught error in iOS.

Differential Revision: D74197282

### DIFF
--- a/torchx/runner/test/api_test.py
+++ b/torchx/runner/test/api_test.py
@@ -153,6 +153,7 @@ class RunnerTest(TestWithTmpDir):
             )
 
             app_handle = runner.run(app, "local", self.cfg)
+            runner.wait(app_handle, wait_interval=0.1)
 
             scheduler, session_name, app_id = parse_app_handle(app_handle)
             self.assertEqual(scheduler, "local")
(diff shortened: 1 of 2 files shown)


highker (approved):

highker (approved):
I just want to try again

PR #1076, opened 2025-06-02
update pyre nightly version

daniel-ohayon:
Summary: Pyre signal is failing on github because the nightly version that we're pinned to is no longer available in pip

Differential Revision: D75791451


kiukchung:
The corresponding internal change (D75791451) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #1080, opened 2025-06-12
(torchx/scheduler) Fill hostnames for each replica in slurm scheduler's describe API

kiukchung:
Summary: Use `scontrol` to implement the describe API that fills out the hostnames for each replica.

Differential Revision: D76485112

### DIFF
--- a/torchx/schedulers/slurm_scheduler.py
+++ b/torchx/schedulers/slurm_scheduler.py
@@ -20,6 +20,7 @@ import subprocess
 import tempfile
 from dataclasses import dataclass
 from datetime import datetime
+from subprocess import CalledProcessError, PIPE
 from typing import Any, Dict, Iterable, List, Mapping, Optional, Tuple
 
 import torchx
@@ -39,6 +40,7 @@ from torchx.specs import (
     macros,
     NONE,
     ReplicaStatus,
+    Resource,
     Role,
     RoleStatus,
     runopts,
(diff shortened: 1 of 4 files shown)


torchx/schedulers/slurm_scheduler.py
```diff
@@ -482,10 +488,82 @@ def _cancel_existing(self, app_id: str) -> None:
...
+            self._describe_scontrol,  # NOTE: only scontrol fills hostnames
+            self._describe_sacct,
+            self._describe_squeue,
+        ]:
+            try:
+                return describe(app_id)
+            except CalledProcessError:
+                continue
+
+    def _describe_scontrol(self, app_id: str) -> Optional[DescribeAppResponse]:
```
d4l3k, line 502: `scontrol show --json job 221` and `squeue --json -j 221` appear to be identical outputs. Can we just merge this into the existing `_describe_squeue` method below?

d4l3k (changes requested):

torchx/schedulers/slurm_scheduler.py
kiukchung: done

torchx/schedulers/slurm_scheduler.py
```diff
@@ -540,46 +562,79 @@ def _describe_sacct(self, app_id: str) -> Optional[DescribeAppResponse]:
...
+                        hostname=hostname,
+                    )
+                )
 
         return DescribeAppResponse(
             app_id=app_id,
             roles=list(roles.values()),
             roles_statuses=list(roles_statuses.values()),
-            state=app_state,
-            msg=msg,
```
d4l3k, line 582: msg isn't needed?

d4l3k (approved):
LGTM

torchx/schedulers/slurm_scheduler.py
kiukchung: yea `msg` defaults to an empty string if not specified. We were just setting `msg=state` so no real functional value added + describe_sacct didn't set msg

PR #1081, opened 2025-06-20
Fix the vulnerability due to protobuf version

li-boyuan:
Differential Revision: D77037884


kiukchung:
The corresponding internal change (D77037884) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #1087, opened 2025-07-01
(torchx/entrypoints) Use importlib.metadata instead of the backported importlib_metadata

kiukchung:
Summary: torchx dropped support for python<=3.7 2 years ago (see: https://github.com/pytorch/torchx/pull/728). But we kept using the backported `importlib_metadata` module to load configurations from `entry_points`.

Differential Revision: D77619564


andywag (approved):
lgtm

PR #1088, opened 2025-07-01
(torchx/components) Remove dependency to pyre_extensions

kiukchung:
Summary: There was only a single usage of `pyre_extensions` in the entire code base. Get rid of this dependency by using an `assert` statement rather than `pyre_extensions.none_throws`.

Reviewed By: highker

Differential Revision: D77621041


andywag (approved):
lgtm

PR #1093, opened 2025-07-23
(torchx/runopt) Allow runopt type to be builtin list[str] and dict[str,str]

ethanbwaite:
Summary:
As title, supports `list[str]` and `dict[str, str]` types in runopt.

Updated testcases to better cover potential type casting issues.

Differential Revision: D78767495

### DIFF
--- a/torchx/specs/api.py
+++ b/torchx/specs/api.py
@@ -789,6 +789,60 @@ class runopt:
     is_required: bool
     help: str
 
+    @property
+    def is_type_list_of_str(self) -> bool:
+        """
+        Checks if the option type is a list of strings.
+
+        Returns:
+            bool: True if the option type is either List[str] or list[str], False otherwise.
+        """
+        return self.opt_type in (List[str], list[str])
+
+    @property
+    def is_type_dict_of_str(self) -> bool:
+        """
+        Checks if the option type is a dict of string keys to string values.
+
+        Returns:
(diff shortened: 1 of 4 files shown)


torchx/specs/api.py
```diff
@@ -789,6 +789,48 @@ class runopt:
     is_required: bool
     help: str
 
+    @property
+    def is_type_list(self) -> bool:
```
tonykao8080, line 793: should the property be renamed to 'is_type_list_of_str' since the check ensure the elements are of type str?

torchx/specs/api.py
```diff
@@ -789,6 +789,48 @@ class runopt:
...
+        return self.opt_type in (List[str], list[str])
+
+    @property
+    def is_type_dict(self) -> bool:
```
tonykao8080, line 797: similar to is_type_list, should this be more explicitly named as type_dict_of_str?

torchx/specs/api.py
ethanbwaite: Agreed, updated

torchx/specs/api.py
ethanbwaite: Same as above, updated

tonykao8080 (approved):
looks good

kiukchung (approved):

PR #1102, opened 2025-08-27
fix: Pass instance type into aws batch job definition even when num_replicas = 1

dataproblems:
Changed the condition for including instance type from `num_replicas > 1` to `num_replicas > 0`
    - In a situation where an aws batch job had multiple roles associated with the nodes, the batch job definition would error out upon submission due to the missing instance type if any of the roles had a single replica associated with it. 

Test plan:

1. Updated existing unit tests to verify that the instance type gets passed. 
2. Created a job using my patch and validated the behavior.


clumsy:
For your consideration, @kiukchung @d4l3k

kiukchung (approved):

PR #1123, opened 2025-09-19
(bugfix)(torchx/components) handle BinOp style optional when decoding component arguments

kiukchung:
Summary:
Fixes https://fb.workplace.com/groups/140700188041197/posts/1025813616196512/?comment_id=1025888432855697&reply_comment_id=1031032659007941

Adds handling for BinOp style (introduced in python-3.10 - **PEP 604**) when decoding component arguments from cli arguments.

For example:

```
def my_component(env: dict[str, str] | None = None) -> AppDef: 
   ...
```

Handles passing `torchx run foo.py:my_component --env FOO:BAR;BAR=BAZ`

Differential Revision: D82856462


ishachirimar (approved):

ethanbwaite (approved):

PR #1137, opened 2025-10-01
support older slurm job_state schema

amirafzali:
Summary:
With SLURM <= 23.02, job_state was a string and not a list. Even with the list we only use the first element.
Added support for string type

https://slurm.schedmd.com/archive/slurm-23.02-latest/rest_api.html#v0_0_39_job_state

Differential Revision: D83690140


kiukchung (approved):

kiukchung:
please fix `torchx/schedulers/slurm_scheduler.py:76:26 Missing parameter annotation [2]: Parameter `job` must have a type other than `Any`. pyre error before landing

kiukchung:
> > please fix `torchx/schedulers/slurm_scheduler.py:76:26 Missing parameter annotation [2]: Parameter `job`must have a type other than`Any`. pyre error before landing
> 
> hm it's a loaded json (https://github.com/meta-pytorch/torchx/blob/main/torchx/schedulers/slurm_scheduler.py#L654-L663), do we have an existing type for this?

pyre doesn't like it when public functions use `Any`. For json repr, you can use `dict[str, object]`

PR #1147, opened 2025-10-16
[torchx][CI] Unpin torch, torchvision, torchxtext to fix breaking uni…

kiukchung:
yesterday (10/15/2025) there was a new release of torch, torchvision, and torchtext.

Usually these three libs are released on the same day and need to be on matching releases to work together. Our `dev-requirements.txt` had torchvision pinned to `0.23.0` but torch pinned (loosely) to `>=2.7.0`. Therefore, as of 10/15, the CI installed `torch-2.9.0` and `torchvision-0.23.0` (not `0.24.0`) hence causing a segfault.

This PR unpins torch, torchvision, and torchtext so that we grab the latest versions of all three packages.

### DIFF
--- a/torchx/specs/api.py
+++ b/torchx/specs/api.py
@@ -18,7 +18,7 @@ import typing
 import warnings
 from dataclasses import asdict, dataclass, field
 from datetime import datetime
-from enum import Enum
+from enum import Enum, IntEnum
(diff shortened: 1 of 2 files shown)


AbishekS (approved):
lgtm

Issue #1150, opened 2025-10-17
torchx status command is broken for missing k8s job


clumsy:
## 🐛 Bug

If job is missing (deleted or not existing) `torchx status` command should not break, but show `State: UNKNOWN` and report the error

Module (check all that applies):
 * [ ] `torchx.spec`
 * [ ] `torchx.component`
 * [ ] `torchx.apps`
 * [ ] `torchx.runtime`
 * [ ] `torchx.cli`
 * [x] `torchx.schedulers`
 * [ ] `torchx.pipelines`
 * [ ] `torchx.aws`
 * [ ] `torchx.examples`
 * [ ] `other`

## To Reproduce

1. `torchx status kubernetes://torchx/real-namespace:bogus-job-id`

```
Traceback (most recent call last):
  File "/opt/conda/bin/torchx", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/torchx/cli/main.py", line 118, in main
    run_main(get_sub_cmds(), argv)
  File "/opt/conda/lib/python3.12/site-packages/torchx/cli/main.py", line 114, in run_main
    args.func(args)
  File "/opt/conda/lib/python3.12/site-packages/torchx/cli/cmd_log.py", line 189, in run
    get_logs(
  File "/opt/conda/lib/python3.12/site-packages/torchx/cli/cmd_log.py", line 99, in get_logs
    status = runner.status(app_handle)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/torchx/runner/api.py", line 491, in status
    desc = scheduler.describe(app_id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/torchx/schedulers/kubernetes_scheduler.py", line 711, in describe
    resp = self._custom_objects_api().get_namespaced_custom_object_status(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/azzhipa/.local/lib/python3.12/site-packages/kubernetes/client/api/custom_objects_api.py", line 1932, in get_namespaced_custom_object_status
    return self.get_namespaced_custom_object_status_with_http_info(group, version, namespace, plural, name, **kwargs)  # noqa: E501
[...]

PR #1151, opened 2025-10-17
fix: handle kubernetes describe failures gracefully (#1150)

clumsy:
When job is not found we will show State=UNKNOWN and surface the error in Msg

Test plan:
[x] added unit test
[x] `torchx status sfai_kubernetes://torchx/real-namespace:bogus-job-id`

```
AppStatus:
    State: UNKNOWN
    Num Restarts: -1
    Roles: 
    Msg: Not Found: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"jobs.batch.volcano.sh \"bogus-jobs-id\" not found","reason":"NotFound","details":{"name":"bogus-job-id","group":"batch.volcano.sh","kind":"jobs"},"code":404}

    Structured Error Msg: <NONE>
    UI URL: None
```


kiukchung:
> Do you mean `State: None`? What would you like it to show @kiukchung ? I'm happy to update my fix as long as it's not an exception.

I mean that if volcano raises an exception when the job does not exist, then torchx's `describe()` API should return `None` as per its documentation: https://meta-pytorch.org/torchx/latest/runner.html

```
describe(app_handle: str)) → Optional[AppDef]
  """
   Reconstructs the application (to the best extent) given the app handle. Note that the reconstructed application may not be the complete app as it was submitted via the run API. How much of the app can be reconstructed is scheduler dependent.

    Returns:

        AppDef or None if the app does not exist anymore or if the scheduler does not support describing the app handle
  """
```

PR #1154, opened 2025-10-21
Introduce StructuredRunOpt for more complex runopts

AbishekS:
Summary:
StructuredRunOpt can be used to create new complex runopts

This can be used as for example:

```
    dataclass
    class UlimitTest(StructuredRunOpt):
        name: str
        hard: int
        soft: int

        def template(self) -> str:
            return "{name},{soft},{hard}"
```
This comes with 
1. template() that helps the from_repr() use that template to map to the fields with types
2. __eq__ to check equality between two

Differential Revision: D85159071


kiukchung:
The corresponding internal change (D85159071) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #1160, opened 2025-11-04
add monarch app def to use monarchs torchrun

colin2328:
Summary:
RFC diff to use monarch's torchrun runner .

creates an app def that

Differential Revision: D86155019


kiukchung:
The corresponding internal change (D86155019) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #1187, opened 2026-01-20
Fix generate version code to generate _version.py instead of version.py

AbishekS:


kiukchung:
The corresponding internal change (D91055760) was abandoned/superseded; closing. Please re-open or re-spin against current main if still relevant.

PR #1208, opened 2026-02-02
fix minikube timeouts

AbishekS:
Differential Revision: D92064020


kiukchung:
The minikube CI job this fixes has been disabled since Feb 2026 with no re-enable plan; closing - happy to revisit if that CI returns.

PR #1253, opened 2026-03-12
Add scoped deprecation warnings for entry-point plugins (#1253)

kiukchung:
Summary:

Add `torchx.deprecations` module with `deprecated_entrypoint()` that scopes
deprecation warnings to only PluginType-defined groups.

**Problem:** Entry-point plugins needed deprecation warnings to guide users
toward the new `torchx_plugins.*` namespace-package system, but the warnings
should only fire for groups that actually have namespace alternatives
(schedulers, named_resources, tracker) — not for every entry-point group
(components, file, schedulers.orchestrator).

**Solution:**

  deprecated_entrypoint(group, ep_names)
      │
      ├── group in _PLUGIN_GROUPS?  → emit DeprecationWarning
      │     (torchx.schedulers, torchx.named_resources, torchx.tracker)
      │
      └── group NOT in _PLUGIN_GROUPS?  → silently return (no warning)

**Changes:**
- New `torchx.deprecations` module with `deprecated_entrypoint()`
- `_PLUGIN_GROUPS` frozenset kept in sync with `PluginType` via test assertion
- `_registry._find()` calls `deprecated_entrypoint()` when entry points are loaded
- Warning message includes migration guidance and `TORCHX_NO_ENTRYPOINTS=1` opt-out

Reviewed By: AbishekS

Differential Revision: D96212551

### DIFF
--- a/torchx/deprecations.py
+++ b/torchx/deprecations.py
@@ -23,7 +23,61 @@ from __future__ import annotations
 
 import functools
 import warnings
-from typing import Any, Callable
+from typing import Any, Callable, Iterable
+
+
+# Entry-point groups that have ``torchx_plugins.*`` namespace-package
+# alternatives.  Only these groups trigger a deprecation warning.
+# Keep in sync with ``torchx.plugins._registry.PluginType``.
+_PLUGIN_GROUPS: frozenset[str] = frozenset(
+    {
+        "torchx.schedulers",
+        "torchx.named_resources",
+        "torchx.tracker",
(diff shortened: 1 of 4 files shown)


kiukchung:
Abandoning the backing diff D96212551 — the scoped entry-point deprecation warnings need re-authoring against the refactored PluginRegistry (PluginSource/TORCHX_PLUGINS_SOURCE) API; not worth carrying. Closing.

PR #1275, opened 2026-04-05
Discover implicit namespace subpackages in plugin scanner (#1275)

AbishekS:
Summary: Pull Request resolved: https://github.com/meta-pytorch/torchx/pull/1275

Differential Revision: D99603190

### DIFF
--- a/torchx/plugins/_registry.py
+++ b/torchx/plugins/_registry.py
@@ -20,6 +20,7 @@ import functools
 import importlib
 import logging
 import os
+import pathlib
 import pkgutil
 from types import ModuleType
 from typing import Any, Callable, overload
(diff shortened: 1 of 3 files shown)


torchx/plugins/_registry.py
```diff
@@ -174,6 +175,12 @@ def _scan_namespace_pkg(
         Public submodules (not starting with ``_``) are imported.  Sub-packages
         are recursed into automatically (e.g. ``msl/``, ``fair/``).
 
+        After ``pkgutil.iter_modules``, also probes for **implicit namespace
+        sub-packages** — directories without ``__init__.py`` that contain
+        ``.py`` files.  This lets namespace packages like
+        ``torchx_plugins.named_resources`` host subdirectories (``mast/``,
```
kiukchung: remove internal references like mast and mkube.

torchx/plugins/test/registry_test.py
```diff
@@ -463,6 +465,15 @@ def test_discovers_named_resources_from_multiple_packages(self) -> None:
         res_gcp = result["gcp_a3_highgpu_8g"]()
         self.assertEqual(res_gcp.cpu, 252, "gcp_a3_highgpu_8g should have 252 CPUs")
 
+        # --- implicit_sub/hardware.py: no __init__.py in implicit_sub/ ---
+        self.assertIn(
+            "implicit_gpu",
```
kiukchung: should be renamed to something more realistic. No one will make a named resource be called "implicit_gpu"

kiukchung (commented):
Can you make sure this won't add significant scan latency in addition to what we already have? you can test in the tbr conda env for a realistic latency measurement. We only scan once and cache.

PR #1287, opened 2026-04-22
Fix CUSTOM_NAMED_RESOURCES priority in _load_named_resources

AbishekS:
Summary: Swap the unpack order so namespace-scanned plugins form the base layer and the in-tree registries (GENERIC/AWS/CUSTOM) override them on name collision. Previously namespace plugins were unpacked last, so a downstream ``torchx_plugins.named_resources.*`` registration silently shadowed ``CUSTOM_NAMED_RESOURCES`` — including any capability injection the custom layer adds on top of the base factory. After: ``CUSTOM_NAMED_RESOURCES`` wins, matching the docstring contract downstream plugin authors rely on.

Differential Revision: D101888085

### DIFF
--- a/torchx/plugins/test/registry_test.py
+++ b/torchx/plugins/test/registry_test.py
@@ -219,6 +219,40 @@ class RegistryTest(_RegistryTestBase):
         )
 
     @mock_install_torchx_plugins()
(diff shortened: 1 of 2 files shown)


kiukchung:
Superseded by the plugin-based resource resolution direction; closing.

PR #1317, opened 2026-06-18
feat: add remaining aws m5 instances types

clumsy:
Adding the remaining m5 instance types according to https://aws.amazon.com/ec2/instance-types/m5

The motivation is to have these built-in CPU only `Resource`s for heterogeneous disaggregated RL training runs.

Full specs: https://docs.aws.amazon.com/ec2/latest/instancetypes/gp.html

Test plan:
- [x] unit tests


clumsy:
Not urgent, but please consider this too @kiukchung @d4l3k @AbishekS

d4l3k (approved):
LGTM

clumsy:
Thanks @d4l3k!

PR #1359, opened 2026-08-14
(cleanup)[torchx][specs] Remove the dead base_img_root macro sentinel

kiukchung:
Summary:
Deletes the long-deprecated `macros.base_img_root` template variable and the `macros.Values.base_img_root = "DEPRECATED"` sentinel field. An fbsource-wide sweep (all 17 `base_img_root` hits accounted for) found **no producer or consumer** outside the definition, its own tests, and one docs bullet: no component references `${base_img_root}` and nothing constructs `Values(base_img_root=...)` outside `specs/test/api_test.py`. `torchx.specs.fb.macros.Values` inherits the field, so the serialized macro-values dict (fed only to `Template.safe_substitute`, where an extra key is inert) drops the `base_img_root` key — `api_meta_test`'s `to_dict()` expectation updated accordingly.

NOTE: the adjacent deprecated `Role.overrides` + `__getattribute__` lazy-resolution machinery is deliberately NOT removed — the APF launcher still produces it at runtime (`apf/launcher/torchx_utils.py` sets `role.overrides` from `PkgInfo.lazy_overrides`, including an async fbpkg-build future that resolves through `Role.__getattribute__` on first `role.image` read).

Differential Revision: D115980859

### DIFF
--- a/docs/source/specs.rst
+++ b/docs/source/specs.rst
@@ -48,14 +48,13 @@ Macros
 
 .. note::
 
-   In addition to the three macros listed in the class docstring, two more
-   attributes exist:
+   In addition to the three macros listed in the class docstring, one more
+   attribute exists:
 
    * ``macros.rank0_env`` -- expands to the **name** of the environment variable
      that provides the rank-0 (master) host address. Resolve it via shell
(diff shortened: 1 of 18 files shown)


seemethere (approved):

PR #1361, opened 2026-08-14
(cleanup)[torchx][schedulers] Delete the aws_batch scheduler (#1361)

kiukchung:
Summary:

Deletes the `aws_batch` scheduler, its test, docs page, and all its wiring: the `DEFAULT_SCHEDULER_MODULES` registration, buck targets (`schedulers:aws_batch_scheduler`, `schedulers/test:aws_batch_scheduler_test`) and their `manual` deps in `lib_noflow`/`runner:lib`/`runner/test`/`github/scripts`, the fb `entry_points.txt` row, the `aws_batch = ["boto3"]` extra plus the `boto3`/`moto` dev deps (re-locked; removals only), the GH Actions `aws_batch --mock` matrix leg (with `_mock_aws_batch` in the integration script), and the dispatch-only disabled `aws-batch-integration-tests.yaml` workflow together with its `awsbatchint.sh`.

Nobody schedules with it: Scuba `pytorch_elastic_tsm_log` (180d) shows only bot traffic on `aws_batch` with zero schedule events, and the only fbsource references were torchx's own wiring above.

Also removes the now-producer-less `ENV_TORCHX_ROLE_IDX`/`ENV_TORCHX_ROLE_NAME` constants from `torchx/settings.py` — only the deleted scheduler set them.

NOTE: `specs/named_resources_aws` stays — `schedulers/devices.py` (EFA/Neuron mounts), `specs/__init__.py`, and specs tests use it beyond `aws_batch`. Out-of-tree string mentions left as-is: `torchrecipes/paved_path` (README + `charnn_dag.py`) and `pytorch/PFNs/.torchxconfig`. `genai/msl` reads the `TORCHX_ROLE_NAME` env var only as an `os.environ.get` fallback behind MAST vars — no import, unaffected.

Differential Revision: D116003521


seemethere (approved):

PR #1362, opened 2026-08-14
(cleanup)[torchx][tracker] Delete the MLflow tracker backend (#1362)

kiukchung:
Summary:

Deletes the MLflow tracker backend (`tracker/mlflow.py` + its test) and the `mlflow-skinny` dev dep that existed only for that test (re-locked; removals only). The tracker *interface* stays: `tracker/api.py`, runner wiring, and `cli/cmd_tracker.py` are untouched. Nothing imports `torchx.tracker.mlflow` — no buck target owns either file, and the only fbsource references were torchx's own docs plus `tracker/test/api_test.py`, which used `MLflowTracker` merely as a mock `spec` in a module-factory-resolution test (now specced on the test's own `TestTrackerBackend`).

NOTE: two items from this cleanup's original scope are deliberately NOT deleted, b/c they have live consumers:

1. `tracker/backend/fsspec.py` — `examples/apps/tracker` (BUCK dep + entry_points) uses it, and the examples sub-project is out of scope here; the `[torchx.tracker] fsspec` entry points (OSS + fb) stay with it.
2. `runtime/tracking/**` — hard buck deps from `ax/metrics` (`//torchx/runtime/tracking:lib` in `ax/metrics/BUCK`, so the `try/except ImportError` in `ax/metrics/torchx.py` does not protect the build graph), `apps/utils:booth`, `specs/test:finder_test`, `github/docs`, and the actively cont-built `bento_kernel_torchx` (v1673, 2026-08-13).

Differential Revision: D116003514


seemethere (approved):

PR #1363, opened 2026-08-14
(cleanup)[torchx][components] Delete the docstring-only train and interpret component stubs (#1363)

kiukchung:
Summary:

Deletes `components/train.py` and `components/interpret.py` plus their docs pages and toctree/table rows. Both modules contain zero code — only a docstring saying "we don't provide an out of the box app, see Captum / `utils.python` instead" — so `torchx builtins` lists nothing from them and nothing imports them (no buck target owns either file; repo search for `components.train`/`components.interpret` after the delete: zero hits outside CHANGELOG/egg-info). The pointers they carried already live in `component_best_practices.rst` and `components/utils`.

Reviewed By: seemethere

Differential Revision: D116003515


seemethere (approved):

PR #1364, opened 2026-08-14
(cleanup)[torchx][components] Delete the serve component and torchx.apps.serve

kiukchung:
Summary:
Deletes the dead torchserve chain: the `serve.torchserve` component (`components/serve.py`), the `torchx.apps.serve` upload app it launched (`apps/serve/**` incl. tests), their buck targets and deps (`export_components`, `apps:penv`, integration-test provider), the `ServeComponentProvider` (its own TODO said the e2e path was never wired), the lightning example's torchserve export chain (`examples/apps/lightning/handler/`, `export_inference_model` + its `torch-model-archiver` shell-out, the trainer's `--skip_export` flag and the integration app-def's use of it), the docs page + toctree/table rows, and the `torchserve`/`torch-model-archiver`/`ts` dev deps (re-locked; removals only).

The component just wrapped `python -m torchx.apps.serve.serve` to POST a model URL to a torchserve management API — nothing in fbsource imports either module. The lightning handler had to go with the deps: OSS pyre flagged its now-unresolvable `ts.torch_handler` import once `torchserve` left the venv.

NOTE: `fbcode/videoinfra/ai_infra/torchx/components/opec_tag.py` keeps a `"serve.torchserve"` row in its `BUILTIN_MAPPING` — a string map resolved lazily via `importlib` only when someone launches that builtin, so it degrades to its existing unknown-builtin `ImportError` handling; left as-is.

Reviewed By: seemethere

Differential Revision: D116003523


seemethere (approved):

PR #1367, opened 2026-08-14
(cleanup)[torchx][github/ci] Delete the dead minikube integration-test workflow (#1367)

kiukchung:
Summary:

Deletes `kubernetes-minikube-integration-test.yaml`. The job has been hard-disabled with `if: false` ("Temporarily disabled") since minikube CI was turned off in D92554147, with no re-enable plan — it triggers on every push/PR only to skip. The kubernetes scheduler itself stays (live internal RMUC consumer); `scripts/setup_minikube.sh` also stays (referenced by the disabled kubernetes leg of `components-integration-tests.yaml`).

Reviewed By: seemethere, athmasagar

Differential Revision: D116003519


seemethere (approved):

PR #1397, opened 2026-08-17
(bugfix)[torchx][plugins] Equal-value same-name plugin re-registration is idempotent (#1397)

kiukchung:
Summary:

When the same plugin is importable through two channels (e.g. installed as a wheel AND present on the source tree), discovery sees it twice — both channels register the same name with the same value:

```python
# portion_a/mod.py AND portion_b/mod.py — same plugin, two install channels
def t_test() -> Resource:
    return Resource(cpu=1, gpu=0, memMB=1024)
```

**Before** — the second registration was recorded as an error, so consumers that assert clean discovery failed on a perfectly healthy install:

```
>>> registry().errors
[RegistrationError(module="portion_b.mod", name="t_test",
    error="duplicate — already discovered, keeping first occurrence")]
```

**After** — an equal-value re-registration is a no-op (first occurrence kept):

```
>>> registry().errors
[]
```

**Root cause:** the duplicate check compared names only, never values, so a benign double-install was indistinguishable from a real conflict. A DIFFERENT value under the same name still records the error (message now says "with a DIFFERENT value") — that is real drift and must stay loud.

NOTE: equality invokes the two factories and compares the materialized resources under structural normalization, because thrift enums/structs regenerated per distribution defeat `==` by class identity even on identical data. `NAMED_RESOURCE` only — factories are cheap and pure; schedulers/trackers are never invoked, so their distinct callables still error.

Differential Revision: D116171124

### DIFF
--- a/torchx/plugins/_registry.py
+++ b/torchx/plugins/_registry.py
@@ -127,6 +127,126 @@ class PluginRegistry:
         """
         return f"torchx_plugins.{pt.value.removeprefix('torchx.')}"
 
+    @staticmethod
+    def _normalize_value(v: Any) -> Any:
+        """Structural normalization for cross-distribution value comparison.
+
+        Two portions of the namespace may carry twin GENERATED types (thrift
+        enums/structs regenerated per distribution) whose ``__eq__`` is
+        class-identity-based — semantically identical values compare unequal.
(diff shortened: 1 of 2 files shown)


seemethere (approved):

PR #1401, opened 2026-08-20
(docs)[4/5][torchx] Document whether dryrun/submit_dryrun mutate the caller's AppDef (#1401)

kiukchung:
Summary:

`Runner.dryrun()` and `Scheduler.submit_dryrun()` disagree on whether they
mutate the `AppDef` you hand them, and neither said so. So reading `role.image`
back off your own `AppDef` after `runner.dryrun()` returns the image you
authored, not the one that will be submitted:

```python
app = AppDef(roles=[Role(image="foo:latest", workspace=..., ...)])
info = runner.dryrun(app, "mast")

app.roles[0].image       # "foo:latest"            <- as authored
info.app.roles[0].image  # "foo:<workspace-hash>"  <- actually submitted
```

That read is what sent 44 multi-region jobs to a stale `:latest`.

| Call | Caller's `AppDef` | `info.app` is |
|---|---|---|
| `Runner.dryrun` / `run` | untouched | a private copy |
| `Scheduler.submit_dryrun` | mutated | the object passed in |
| `Scheduler.submit` | mutated | n/a |

Four more aliasing/defaulting surprises get the same treatment, in the
docstrings rather than here: shared `Role.overrides`, `run_opts()` defaults in
`info.cfg`, `role.workspace` surviving a build, and `dryrun`'s `ValueError`s.

`_submit_dryrun` names the paths that guarantee its roles arrive built, rather
than claiming they always do — a direct `submit_dryrun` caller can still pass an
unbuilt `role.workspace`.

Examples are `.. code-block::` rather than `.. doctest::` because they need a
live scheduler.

Reviewed By: athmasagar

Differential Revision: D116718390

### DIFF
--- a/docs/source/workspace.rst
+++ b/docs/source/workspace.rst
@@ -63,6 +63,22 @@ When ``workspace=`` is passed to :py:meth:`~torchx.runner.Runner.run` (or
    :py:meth:`~torchx.workspace.WorkspaceMixin.push_images` handle pushing
    the built image to a remote registry.
 
+.. important::
+
+   The role mutated in step 3 belongs to a **copy** of your
+   :py:class:`~torchx.specs.AppDef` that
+   :py:meth:`~torchx.runner.Runner.dryrun` takes before patching -- your own
+   ``AppDef`` still carries the image you authored. To see what was actually
+   submitted, read :py:attr:`~torchx.specs.AppDryRunInfo.app`:
+
+   .. code-block:: python
+
+      info = runner.dryrun(app, "kubernetes", cfg)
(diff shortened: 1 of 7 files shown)


kiukchung:
Closing: the content is already on `main`.

Every file in this PR matches `main` or is behind it. Diffing each of the seven
files at this branch against `origin/main` (`2bd4d643`) returns zero lines that
exist here and not there; `torchx/runner/api.py` here is the older revision,
missing `build_workspace` and `describe_native`. The docstrings landed with
16de100e (#1402), imported from D116718390 on 2026-08-25, and the branch now
conflicts with `main`.

Nothing to merge, so closing rather than rebasing.

